PR #129, opened 2021-12-21
Utils for version

pushpender-singh-ap:
Now you can manage the version through one file.


rajarsheechatterjee:
Removed useKeepAwake from Main.js. You can update the previous PR with the changes.

PR #173, opened 2022-02-23
TeamXNovel

yemenihandsome:
Closes LNReader/lnreader-sources#175

### DIFF
--- a/src/sources/multisrc/madara/MadaraGenerator.js
+++ b/src/sources/multisrc/madara/MadaraGenerator.js
@@ -1,4 +1,4 @@
-import MadaraScraper from './MadaraScraper';
+7import MadaraScraper from './MadaraScraper';
 
 const getPath = extensionId => {
   const path = {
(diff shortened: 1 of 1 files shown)


src/sources/multisrc/madara/MadaraGenerator.js
```diff
@@ -1,4 +1,4 @@
-import MadaraScraper from './MadaraScraper';
+7import MadaraScraper from './MadaraScraper';
```
rajarsheechatterjee, line 1: Remove `7`

src/sources/multisrc/madara/MadaraGenerator.js
```diff
@@ -257,3 +257,11 @@ export const Novel4UpScraper = new MadaraScraper(
   false,
   2,
 );
+export const TeamXNovelScraper = new MadaraScraper(
+  104,
```
rajarsheechatterjee, line 261: `104` is already assigned to a different source. Check for the id in `src/sources/sources.json` and add the source in that file too.

rajarsheechatterjee: Also add `TeamXNovelScraper ` in `src/sources/sources.js`

PR #274, opened 2022-05-27
feat(webview-scroll): webview scrolls to saved

SkillGG:
Now webview has webViewScroll which after any change scrolls webView to given percentage

### DIFF
--- a/src/screens/reader/components/WebViewReader.js
+++ b/src/screens/reader/components/WebViewReader.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useRef, useEffect } from 'react';
 import { StatusBar, StyleSheet, View } from 'react-native';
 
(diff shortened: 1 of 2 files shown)


rajarsheechatterjee (approved):

src/screens/reader/components/WebViewReader.js
```diff
@@ -12,14 +12,32 @@ const WebViewReader = ({
...
+  webViewScroll,
 }) => {
   const backgroundColor = readerBackground(reader.theme);
 
   const chapterText = sanitizeChapterText(html);
 
+  const webViewRef = useRef(null);
+
+  useEffect(() => {
+    if (webViewRef.current && webViewScroll > 0) {
```
SkillGG, line 25: remove webViewScroll >= 0 should be used to be able to scroll to top!

PR #354, opened 2022-08-07
handle devices with a notch when reading

HZSamir:
When reading the text used to go behind the notch on devices that have them.
Not anymore.

### DIFF
--- a/src/screens/reader/ReaderScreen.js
+++ b/src/screens/reader/ReaderScreen.js
@@ -9,6 +9,7 @@ import {
   NativeModules,
   NativeEventEmitter,
 } from 'react-native';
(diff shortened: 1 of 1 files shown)


rajarsheechatterjee:
I think this should be an optional setting as I personally don't mind text in the cutout area.

You can add a setting `Show content in cutout area` in bottom sheet and set this behaviour true by default.

Axillman:
I love this app but when i try to scroll on a chapter the menu pops up and makes me unable to do so. 
Can you please add a double tap to show the menu. Thank you.

rajarsheechatterjee:
Closing this pull request as it has become stale, and the development branch has been switched to master.

PR #457, opened 2022-10-26
fix date

Rider21:
fixed a typo in the translation (non-existent format)
fixed the display of illustrations in ranobelib sources

### DIFF
--- a/strings/languages/ru/strings.json
+++ b/strings/languages/ru/strings.json
@@ -146,10 +146,10 @@
     "calendar": {
       "sameDay": "[Сегодня]",
       "nextDay": "[Завтра]",
-      "nextWeek": "дддд",
+      "nextWeek": "dddd",
       "lastDay": "[Вчера]",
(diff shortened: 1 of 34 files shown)


Rider21:
That's the typo I meant in the translation.
Screenshot_20221226-131934

strings/languages/ru/strings.json
```diff
@@ -146,10 +146,10 @@
     "calendar": {
       "sameDay": "[Сегодня]",
       "nextDay": "[Завтра]",
-      "nextWeek": "дддд",
+      "nextWeek": "dddd",
```
rajarsheechatterjee, line 149: Since these should not be translated they should be removed from the `strings/languages/en/strings.json` file.

Rider21: I suppose to delete `sameElse` as well ?

rajarsheechatterjee (approved):

PR #476, opened 2022-11-21
Feature/show chapter name

AustriaGamer:
Allows the user to display the title of the chapter above the chapter text or to disable the title.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4138,7 +4138,6 @@
       "version": "0.66.15",
       "resolved": "https://registry.npmjs.org/@types/react-native/-/react-native-0.66.15.tgz",
       "integrity": "sha512-qNK3LZhNpSd2Hdr0OFdvQ/VOb1qcjNhmw96LsOtxIN+1RtXHO/1dVzxYUlWnKMCtV+E5WSaWM703c9Q7gG9tyg==",
-      "dev": true,
       "dependencies": {
         "@types/react": "*"
       }
@@ -18610,7 +18609,8 @@
     "@jsamr/react-native-li": {
       "version": "2.3.1",
       "resolved": "https://registry.npmjs.org/@jsamr/react-native-li/-/react-native-li-2.3.1.tgz",
-      "integrity": "sha512-Qbo4NEj48SQ4k8FZJHFE2fgZDKTWaUGmVxcIQh3msg5JezLdTMMHuRRDYctfdHI6L0FZGObmEv3haWbIvmol8w=="
+      "integrity": "sha512-Qbo4NEj48SQ4k8FZJHFE2fgZDKTWaUGmVxcIQh3msg5JezLdTMMHuRRDYctfdHI6L0FZGObmEv3haWbIvmol8w==",
+      "requires": {}
     },
     "@native-html/css-processor": {
       "version": "1.11.0",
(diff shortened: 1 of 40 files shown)


src/screens/reader/utils/sanitizeChapterName.ts
```diff
@@ -0,0 +1,53 @@
...
+): string => {
+  let toDeleteText = '';
+  for (let i = 0; i <= nameOnLine; i++) {
+    toDeleteText += lines[i];
+  }
+  return text.replace(toDeleteText, '');
+};
+const findChapterName = (lines: Array<string>, chapterName: string) => {
+  for (let index = 0; index < lines.length; index++) {
+    if (lines[index].includes(chapterName)) {
```
rajarsheechatterjee, line 40: Wouldn't passing the entire text and checking if it includes chapterName be same instead of iterating of the lines?

Also, would only checking if the text includes the name be right?

strings/languages/af/strings.json
```diff
@@ -84,7 +84,8 @@
       "showBatteryAndTime": "Show battery and time",
       "showProgressPercentage": "Show progress percentage",
       "swipeGestures": "Swipe left or right to navigate between chapters",
-      "allowTextSelection": "Allow text selection"
+      "allowTextSelection": "Allow text selection",
+      "showChapterName": "Show chapter name"
```
rajarsheechatterjee, line 88: You just need to add the string to `strings/languages/en/strings.json`, Crowdin would automatically add it to rest of the files.

package-lock.json
```diff
@@ -18610,7 +18609,8 @@
     "@jsamr/react-native-li": {
       "version": "2.3.1",
       "resolved": "https://registry.npmjs.org/@jsamr/react-native-li/-/react-native-li-2.3.1.tgz",
-      "integrity": "sha512-Qbo4NEj48SQ4k8FZJHFE2fgZDKTWaUGmVxcIQh3msg5JezLdTMMHuRRDYctfdHI6L0FZGObmEv3haWbIvmol8w=="
+      "integrity": "sha512-Qbo4NEj48SQ4k8FZJHFE2fgZDKTWaUGmVxcIQh3msg5JezLdTMMHuRRDYctfdHI6L0FZGObmEv3haWbIvmol8w==",
```
rajarsheechatterjee, line 18612: Why are there changes in this file?

src/screens/reader/utils/sanitizeChapterName.ts
AustriaGamer: It is possible to pass the entire text, however often the Autor puts the title somewhere inside the text, that is why I only take the first few lines. As to why I separate the name, in one of my test chapters the title had a "-" in between the number and the actual name and in the chapter was the title without the "-" resulting in two titles.

package-lock.json
AustriaGamer: I have no idea, I didn't even realise it.

SkillGG:
From what I see this PR modifies the chapterSanitize function to alter every source and every chapter the same way?
I think it would be wiser to allow sources to parse chapter names by themselves (and remove them from chapter texts) instead of chapterSanitize doing the hard work. For example a lot of novels have chapter titles that are just lines inside the text that someone says/thinks. So removing them may cause the reader to not understand what is happening because we removed an entire line from the text.
The option to add chapter name with chapter number is a good one, but I think it should be done in the "showing" portion of the app (ReaderScreen) rather than in "chapterParsing" portion.

It should be in source creator's best interest to conform with the standard and remove the titles from chapterText and give us the chapterName with chapterText. Our job should be to provide a standard they can conform to and help them in doing so by changing the faulty sources so that they conform to it.

AustriaGamer:
Your look on the PR is from how I understand it correct. It would definitely make more sense to alter the chapters at the sources part and I have to confess, when I started creating this option I was just browsing through the code, found chapterSanitize and thought this would be an appropriate place. I didn't even think about altering it at the source.

However to alter every source takes a lot of time and the reward isn't that much. Additionally my solution worked perfectly, at least with the sources/novels/chapters I tested.

rajarsheechatterjee:
> The option to add chapter name with chapter number is a good one, but I think it should be done in the "showing" portion of the app (ReaderScreen) rather than in "chapterParsing" portion.

> It should be in source creator's best interest to conform with the standard and remove the titles from chapterText and give us the chapterName with chapterText. Our job should be to provide a standard they can conform to and help them in doing so by changing the faulty sources so that they conform to it.

I agree with this, instead of adding it to the chapter text we could use this method to update the chapter name in the header if it's missing or incorrect.

AustriaGamer:
Thank you for the input. I am closing this PR, will rework it and hopefully be able to create another one in a few weeks or so.

However in reference to the last comment, I want to say that the reason why I want the title inside the chapter text is the functionality and aesthetics. As you can see in the pictures below, if a chapter has a long title, it won't fit in the header.

drawing   ▶️                    drawing

PR #481, opened 2022-11-25
fix(delete_chapter): Excessive number of pending callbacks warn

SkillGG:
Make deleting chapters do 5 coccurent deletion at a time, thus removing the `Excessive number of pending callbacks` warn.


rajarsheechatterjee (approved):
Ideally we should use a single SQLite query to delete multiple records.

SkillGG:
Oh yeah. I forgot something like that was an option. I learned fuck ton of SQL in school. I'm going to try and do that tomorrow and see if it works

PR #485, opened 2022-12-04
Feature/chapter drawer

CD-Z:
# Information
This feature adds a drawer to the ReaderScreen which allows the user to see chapters and switch between them.
<div align="center">
chapter drawer - Dark Theme               chapter drawer - Light Theme</div>

## Additional package
- <a href="https://reactnavigation.org/docs/drawer-based-navigation/">@react-navigation/drawer: "^6.0.1"</a>

## Problem
- If Render in HTML is activated, it is not possible to swipe the drawer open, because it doesn't register the swipe on the WebView. 
  - Through changing the padding-left of the html body to a marginLeft of the WebView, it creates a small part where the drawer can be opened.


SkillGG:
Wouldn't activating it by clicking a button be better? Because swiping is reserved for changing to next chapter?

AustriaGamer:
I thought the same, however the swipe gesture of the drawer superimpose over the one of the gesture recognizer and since it is limited to 60 on the left side it is easy to differentiate between the both of them.
The only problem is with the WebView since there is only a very small space where you can drag the drawer.

As for why I didn't use a button. I simply haven't found a place I find suitable and think it is more intuitive to swipe.

SkillGG:
What about there:

AustriaGamer:
No problem. I welcome other opinions and I can't complain to you since I myself am also very nitpicky sometimes. I changed it as yous suggested and left the button as a permanent in there. The setting to enable the drawer now works as a swipeGesture enabler. I really like to be able to swipe to open the drawer, so I really want to let it in there. 
In my opinion a FAB isn't really suitable, since it is to distracting while reading.

rajarsheechatterjee (changes requested):

src/screens/reader/ReaderScreen.js
```diff
@@ -353,10 +382,7 @@ const Chapter = ({ route, navigation }) => {
 
   const onWebViewNavigationStateChange = async ({ url }) => {
     if ((sourceId === 50 || sourceId === 62) && url !== 'about:blank') {
-      setLoading(true);
-      const res = await fetchChapter(sourceId, novelUrl, url);
-      setChapter(res);
-      setLoading(false);
```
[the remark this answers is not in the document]
AustriaGamer: Sorry about that, I don't know why I did this however I reverted the change

AustriaGamer:
In the latest commits I added a ts file, which helps to pass props for navigation.navigate.
I would appreciate feedback if this makes sense or is just unnecessary. Also is the placement of the file good?

src/screens/reader/utils/handleNavigateChapter.ts
```diff
@@ -0,0 +1,67 @@
...
+interface selectFewerProps {
+  chapterId: number;
+  chapterName: string;
+  chapterUrl: string;
+}
+const selectFewerProps = (obj: selectFewerProps) => {
+  const { chapterId, chapterName, chapterUrl } = obj;
+  return { chapterId, chapterName, chapterUrl };
+};
+export function DopenChapter(
```
rajarsheechatterjee, line 33: Where's this function being used?

rajarsheechatterjee:
> In the latest commits I added a ts file, which helps to pass props for navigation.navigate. I would appreciate feedback if this makes sense or is just unnecessary. Also is the placement of the file good?

Yeah, it makes sense to define it at a single place than creating the params object everywhere.

src/screens/reader/utils/handleNavigateChapter.ts
AustriaGamer: Nowhere, I just used it for testing some things. This commit was just to see the reaction on the new function and wasn't finished yet. So I still haven't deleted all unnecessary code.

rajarsheechatterjee: Cool, you could mark the PR as Draft till it's ready.

AustriaGamer: Sure, completely forgot about the existence of drafts.

src/screens/reader/components/ChapterDrawer.js
```diff
@@ -0,0 +1,229 @@
...
+import { getString } from '@strings/translations';
+import { useNovel } from '@hooks/reduxHooks';
+
+const ChapterDrawer = ({ state, navigation }) => {
+  const theme = useTheme();
+  const insets = useSafeAreaInsets();
+  const styles = createStylesheet(theme, insets);
+
+  let listRef = useRef();
+  const { chapters } = useNovel();
```
rajarsheechatterjee, line 19: Chapters is set when the User open the Novel Details screen, what's the behaviour when they navigate from other screens?

src/screens/library/components/LibraryListView.tsx
```diff
@@ -45,10 +46,12 @@ export const LibraryView: React.FC<Props> = ({
         if (selectedNovelIds.length) {
           setSelectedNovelIds(xor(selectedNovelIds, [item.novelId]));
         } else {
-          navigate('Novel' as never, item as never);
+          navigate('Novel' as never, openNovel(item) as never);
           dispatch(setNovel(item));
         }
       }}
+      libraryStatus={undefined}
+      selectedNovels={undefined}
```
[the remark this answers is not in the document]
CD-Z: Removed. I only put them there because the lint plugin was bugging me.

src/screens/reader/components/ChapterDrawer.js
CD-Z: Problem is now resolved

Now, if you navigate from other screens it dispatches setNovel and getNovelAction. It works that way, however I don't think it is the best solution, but the best solution I could come up with.

Also it now adopts the list preferences from the novelscreen. (sort and filter)

rajarsheechatterjee (approved):

PR #492, opened 2022-12-10
chapter drawer themed textcolor

CD-Z:
@rajarsheechatterjee 
First of all I really like your redesign, however you forgot to theme the text color, so it was barely readable in darkmode


rajarsheechatterjee:
Yeah, hadn't worked on the app in a while so forgot about it.

PR #497, opened 2022-12-14
Feature/adaptive icon

CD-Z:
I created new adaptive Icons and removed the round icon definition since it is useless.
If you wonder why the icons are so big, this is because the website I used to create them gave them back in this size. I don't think this creates any problems.

image


rajarsheechatterjee:
@CD-Z The icon has always looked a bit blurry on some devices, should I provide a higher resolution foreground image or convert it to a SVG?

CD-Z:
@rajarsheechatterjee That would be great. Especially on opening the app you see the app icon before the splash screen and it is very pixilated.

rajarsheechatterjee:
Here's a higher resolution [image](https://drive.google.com/file/d/1LLmRQXN-9-cRKTKooKYDQWlugkZacJvj/view?usp=sharing) of the logo.

rajarsheechatterjee (approved):

PR #499, opened 2022-12-18
Feature/animated splashscreen

CD-Z:
I try to create an animated splashscreen. However I can't seem to find a way to center it. Help and thoughts on this would be appreciated.


src/navigators/Main.js
```diff
@@ -41,7 +40,6 @@ const MainNavigator = () => {
 
   useEffect(() => {
     const timer = setTimeout(async () => {
```
rajarsheechatterjee, line 42: You could remove this from `setTimeout` and see if the status bar color is being set during the launch itself.

CD-Z:
Thank you for the suggestion. While it didn't solve the problem, I still found a way to apply the correct Style.

rajarshee-harness (approved):

PR #509, opened 2023-01-05
feat: chapter title options

CD-Z:
This is the improved version of my "show chapter name" PR.
In this version every chapter gets saved with an additional prop chapterPrefix.
- chapterPrefix is a string with the Volume number, a space between and the chapter number.
- chapterName is only the name of the individual chapter or an empty string.

The chapterTitle gets created with the help of hooks because this is the only way I managed to do it without bugs.
Since this PR modifies the db I created a method with an button in the advanced tab of the settings to transform the already existing db to work properly. However it would be better if this method would be called once after the update and restart the app automatically but it couldn't find a way to implement this.


rajarsheechatterjee:
These seem like lot of changes for something which is not that important as I don't see much benefits of having this. Also I don't think saving the prefix in the DB would be a good idea.

CD-Z:
Well I kind of expected this. Originally I wanted to do it a lot simpler too, but it kinda blew up and I just wanted to make it work. For now I will scrap this idea, maybe I will pick it up in the future again.

PR #513, opened 2023-01-09
fixed library button color

CD-Z:
A fix for some dark themes:
- Before
Screenshot_1673287752
- After
Screenshot_1673289705

### DIFF
--- a/src/navigators/BottomNavigator.js
+++ b/src/navigators/BottomNavigator.js
@@ -32,11 +32,7 @@ const BottomNavigator = () => {
           .string(),
       }}
       theme={{ colors: theme }}
-      activeColor={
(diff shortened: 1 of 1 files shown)


src/navigators/BottomNavigator.js
```diff
@@ -32,7 +32,11 @@ const BottomNavigator = () => {
           .string(),
       }}
       theme={{ colors: theme }}
-      activeColor={theme.primary}
+      activeColor={
```
rajarsheechatterjee, line 35: According to Material 3 guidelines, `activeColor` should be On secondary container.

rajarsheechatterjee (approved):

PR #535, opened 2023-02-09
fix slider displays uncorrectly, rearrange seekbar props

nyagami:
slider inside Seekbar component

### DIFF
--- a/src/screens/reader/ReaderScreen.js
+++ b/src/screens/reader/ReaderScreen.js
@@ -135,6 +135,7 @@ const ChapterContent = ({ route, navigation }) => {
   });
   const [firstLayout, setFirstLayout] = useState(true);
   const [scrollPage, setScrollPage] = useState(null);
+
   useEffect(() => {
     VolumeButtonListener.disconnect();
     if (useWebViewForChapter && wvUseVolumeButtons) {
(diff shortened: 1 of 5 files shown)


nyagami:
This fix will reduce number of rerendering of ReaderScreen from 12 to 9. 
Also screen will loading until fetch pre and next chapter

nyagami:
Havent tested in special cases. But I think the seekbar can work quite well from this xD

src/screens/reader/ReaderScreen.js
```diff
@@ -216,6 +204,18 @@ const ChapterContent = ({ route, navigation }) => {
     }
   }, []);
 
+  const [[nextChapter, prevChapter], setAdjacentChapter] = useState([]);
+
+  useEffect(() => {
+    const setPrevAndNextChap = async () => {
+      const nextChap = await getNextChapter(novelId, chapterId);
+      const prevChap = await getPrevChapter(novelId, chapterId);
```
rajarsheechatterjee, line 212: We could use `Promise.all()` for concurrency.

src/screens/reader/components/TextReader.tsx
```diff
@@ -82,7 +82,7 @@ const TextReader: React.FC<TextReaderProps> = ({
               title={`Next: ${nextChapter.chapterName}`}
               onPress={navigateToNextChapter}
               theme={theme}
-              textColor={theme.colorButtonText}
+              textColor={theme.colorButtonText} // what is this prop ?
```
rajarsheechatterjee, line 85: To override the button's text color.

nyagami: but I didnt see colorButtonText in that interface

rajarsheechatterjee: yeah, this was removed when themes were migrated to Material You, and I missed removing this.

rajarsheechatterjee (approved):

rajarsheechatterjee:
Tested the build on emulator, the scroll position is being reset to top in `Render HTML` mode after scrolling to the actual position.

nyagami:
I was quite confused about this thing. Maybe I should add (content size > 0) condition to doSaveProgress()

PR #538, opened 2023-02-14
Merge functions, improve seekbar

nyagami:


src/screens/reader/ReaderScreen.js
```diff
@@ -392,16 +397,7 @@ const ChapterContent = ({ route, navigation }) => {
...
-            swipeGestures &&
-            (!useWebViewForChapter || !wvUseNewSwipes) &&
-            navigateToPrevChapter
-          }
-          onSwipeLeft={
-            swipeGestures &&
-            (!useWebViewForChapter || !wvUseNewSwipes) &&
-            navigateToNextChapter
-          }
+          onSwipe={navigateToChapterBySwipe}
```
rajarsheechatterjee, line 400: `swipeGestures && (!useWebViewForChapter || !wvUseNewSwipes)` should be true for Swipe Navigation to be enabled.

nyagami: Why we cant use `swipeGestures` for `webview`. Also I dont know what `wvUseNewSwipes` does.
So only swipeGestures is enough imo

nyagami: So you should be unnecessary right now?. Cause we can swipe in Web mode without it

src/screens/reader/ReaderScreen.js
```diff
@@ -392,16 +397,7 @@ const ChapterContent = ({ route, navigation }) => {
...
-            swipeGestures &&
-            (!useWebViewForChapter || !wvUseNewSwipes) &&
-            navigateToPrevChapter
-          }
-          onSwipeLeft={
-            swipeGestures &&
-            (!useWebViewForChapter || !wvUseNewSwipes) &&
-            navigateToNextChapter
-          }
+          onSwipe={swipeGestures && !wvUseNewSwipes && navigateToChapterBySwipe}
```
rajarsheechatterjee, line 400: Render HTML and `wvUseNewSwipes` both should be true or only `swipeGestures`.

src/screens/reader/ReaderScreen.js
rajarsheechatterjee: Then we would also have to remove it from the setting, since we are combining them.

nyagami: okay, lemme try

src/redux/settings/settings.reducer.js
```diff
@@ -63,17 +63,14 @@ export const initialState = {
   fullScreenMode: true,
   swipeGestures: false,
   showScrollPercentage: true,
-  useWebViewForChapter: false,
-  wvUseNewSwipes: false,
-  wvShowSwipeMargins: true,
-  wvUseVolumeButtons: false,
+  ShowSwipeMargins: true,
```
rajarsheechatterjee, line 66:
```suggestion
  showSwipeMargins: true,
```

rajarsheechatterjee (approved):

PR #545, opened 2023-02-15
Feature/fix sources

CD-Z:
fixed most broken english sources and removed dead ones


src/sources/en/LightNovelReader.ts
```diff
@@ -155,41 +155,35 @@ const parseChapter = async (novelUrl: string, chapterUrl: string) => {
...
+}
 
-  const result = await fetch(url, { method: 'POST', body: formData });
-  const body = await result.text();
+const searchNovels = async (searchTerm: string) => {
+  const url = `${baseUrl}/search/autocomplete?dataType=json&query=${searchTerm}`;
 
-  const loadedCheerio = cheerio.load(body);
+  const result = await fetch(url, { method: 'POST' });
+  const body = await JSON.parse(await result.text());
```
rajarsheechatterjee, line 169:
```suggestion
  const body = await result.json();
```

CD-Z: Thanks, didn't know this existed.

PR #548, opened 2023-02-16
Feature/settings and translations

CD-Z:
Adds settings and translation strings in generalsettings and readersettings


src/screens/settings/SettingsGeneralScreen/SettingsGeneralScreen.tsx
```diff
@@ -0,0 +1,343 @@
...
+  const {
+    displayMode = DisplayModes.Comfortable,
+    novelsPerRow = 3,
+    showDownloadBadges = true,
+    showNumberOfNovels = false,
+    showUnreadBadges = true,
+    sortOrder = LibrarySortOrder.DateAdded_DESC,
+  } = useLibrarySettings();
+
+  const transformSortOrder = (text: string) => {
```
rajarsheechatterjee, line 46: You can instead create a Object like
```
const sortOrderNameMap = {
  'lastReadAt': 'libraryScreen.bottomSheet.sortOrders.lastRead',
  ...
}
```

and use it like `getString(sortOrderNameMap[sortOrder])`

src/screens/settings/SettingsGeneralScreen/SettingsGeneralScreen.tsx
```diff
@@ -0,0 +1,343 @@
...
+      res.push('Unread');
+    }
+    if (showNumberOfNovels) {
+      if (res.length !== 0) {
+        res.push(', ');
+      }
+      res.push('Number of Items');
+    }
+    return res;
+  };
```
rajarsheechatterjee, line 102:
```suggestion
  const generateNovelBadgesDescription = () => {
    let res = [];
    if (showDownloadBadges) {
      res.push('Download');
    }
    if (showUnreadBadges) {
      res.push('Unread');
    }
    if (showNumberOfNovels) {
      res.push('Number of Items');
    }
    return res.join(', ');
  };
```

src/screens/settings/SettingsGeneralScreen/SettingsGeneralScreen.tsx
CD-Z: Way better solution, however I used the new Map() function for ts reason

PR #553, opened 2023-02-17
refactor swipeGesture, remove swipeMargin

nyagami:
should be changed later if swipeListener wont pleasure users


nyagami:
thank god, no conflict anymore

PR #560, opened 2023-02-22
fix: Sakuranovel

K1ngfish3r:
fixes https://github.com/LNReader/lnreader-sources/issues/456#issue-1593031462

images don't appear consistently, but I think that's more on the source's side.

### DIFF
--- a/src/sources/multisrc/wpmangastream/WPMangaStreamScraper.js
+++ b/src/sources/multisrc/wpmangastream/WPMangaStreamScraper.js
@@ -91,12 +91,12 @@ class WPMangaStreamScraper {
     novel.genre = loadedCheerio('.genxed').text().trim().replace(/\s/g, ',');
 
     novel.summary = loadedCheerio('.entry-content')
-      .find('h3')
+      .find('h3 , p.a')
       .remove()
       .end()
       .prop('innerHTML')
-      .replace(/(<.*?>)/g, '\n')
-      .replace(/(&.*;)/g, '');
+      .replace(/(<.*?>)/g, ' ')
+      .replace(/(&.*;)/g, '\n');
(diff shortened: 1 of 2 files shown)


src/sources/multisrc/wpmangastream/WPMangaStreamScraper.js
```diff
@@ -91,12 +91,12 @@ class WPMangaStreamScraper {
     novel.genre = loadedCheerio('.genxed').text().trim().replace(/\s/g, ',');
 
     novel.summary = loadedCheerio('.entry-content')
-      .find('h3')
+      .find('h3 , p.a')
```
rajarsheechatterjee, line 94: I checked on the source sites, wouldn't `div[itemprop="description"]` be a better selector?

```
novel.summary = loadedCheerio('div[itemprop="description"]').text()
```

K1ngfish3r: I basically want to make the summary look cleaner by removing those within the box
Screenshot from 2023-02-22 12-14-33

so with this
``novel.summary = loadedCheerio('div[itemprop="description"]').text()``
it looks like this
Screenshot from 2023-02-22 12-26-16

so I try to rectify it (somewhat, I wish I could do better) with this,
`` novel.summary = loadedCheerio('.entry-content')
      .find('h3 , p.a')
      .remove()
      .end()
      .prop('innerHTML')
      .replace(/(<.*?>)/g, ' ')
      .replace(/(&.*;)/g, '\n');``
      
so it looks something like this,
Screenshot from 2023-02-22 12-30-15

rajarsheechatterjee (approved):

PR #562, opened 2023-02-22
Feature/expendable list for Updates and Downloads

CD-Z:
Added expandable list for Updates since that way it is easier to check which novels were updated.

Changed navigateToNovel function from tap on image to onLongPress of the list since I couldn't find an other way. (If I used an absolute positioned cover it would disappear after reload.)

Added the expendable list to Download since it was possible that way to remove unnecessary files.


rajarsheechatterjee:
Shouldn't we directly show the update item if there's a single chapter? This would just increase number of clicks for that case.

CD-Z:
Oh right, actually wanted to do this but forgot. I will add it shortly.

CD-Z:
updated:
Screenshot_1677426084

rajarsheechatterjee (approved):

PR #587, opened 2023-03-30
extending Volume button feature, by scroll amount

r1di:
Added volume button scroll amount

### DIFF
--- a/android/app/src/main/java/com/rajarsheechatterjee/VolumeButtonListener/VolumeButtonListener.java
+++ b/android/app/src/main/java/com/rajarsheechatterjee/VolumeButtonListener/VolumeButtonListener.java
@@ -1,11 +1,11 @@
 package com.rajarsheechatterjee.VolumeButtonListener;
-
 import androidx.annotation.NonNull;
 
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReactContextBaseJavaModule;
 import com.facebook.react.bridge.ReactMethod;
+import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.WritableMap;
 import com.facebook.react.modules.core.DeviceEventManagerModule;
 
@@ -27,9 +27,10 @@ public class VolumeButtonListener extends ReactContextBaseJavaModule {
 
     public static boolean prevent = false;
 
-    public static void sendEvent(boolean up) {
+    public static void sendEvent(boolean up, int amount) {
         if (active) {
             WritableMap args = Arguments.createMap();
+            args.putInt("scrollAmount", amount); // add scroll amount to event data
             if (up)
                 appContext.getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class).emit("VolumeUp", args);
             else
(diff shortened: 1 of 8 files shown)


android/app/src/main/java/com/rajarsheechatterjee/VolumeButtonListener/VolumeButtonListener.java
```diff
@@ -27,22 +27,33 @@ public static boolean isActive() {
...
+    private static int scrollAmount = 100;
+
+    public static void setScrollAmount(int amount) {
+        scrollAmount = amount;
     }
 
-    public static void up() {
-        sendEvent(true);
+    public static int getScrollAmount() {
+        return scrollAmount;
```
nyagami, line 56: Why we have to send event with amount value

r1di: the parameter "amount" is the value representing the amount of scrolling desired when either the volume up or volume down button is pressed.

in the MainActivity class, the dispatchKeyEvent() method listens for volume button presses (up and down) and calls the respective VolumeButtonListener.up() and VolumeButtonListener.down() methods with the scroll amount obtained from VolumeButtonListener.getScrollAmount()

nyagami: Actually the value of `scrollAmout` is stored and persisted by `Redux` (when you call dispatch to settingReducer).
You could remove old eventListener and add the new one when `scrollAmout` get changed.
No need to modify the naitve Module. Anyway I think your logic is overhandling
image

r1di: Thanks, I will look into it. 
Never worked before with Redux/JavaScript, gotta use more chadGPT for knowledge gaps😂👌

src/screens/settings/SettingsReaderScreen/SettingsReaderScreen.tsx
```diff
@@ -196,6 +195,28 @@ const SettingsReaderScreen = () => {
...
           theme={theme}
         />
+        {useVolumeButtons ? (
+          <View style={styles.scrollAmount}>
+            <Text style={[labelStyle, styles.paddingRightM]} numberOfLines={2}>
+              {getString('readerScreen.bottomSheet.scrollAmount')}
+            </Text>
+            <TextInput
+              style={{
+                paddingHorizontal: 16,
```
r1di, line 205: could have used a StyleSheet i guess?

android/app/src/main/java/com/rajarsheechatterjee/VolumeButtonListener/VolumeButtonListener.java
r1di: i pushed a new commit, not sure if this was the correct way(im coming from gerrit, is this like a new patchset?).
I think its cleaner now, now working with redux for all the changes.

I tried to use **KeyboardAvoidingView**, but didnt manage to implement it, would be usefull for the future.

thank you for patience😁

rajarsheechatterjee (approved):

android/app/src/main/java/com/rajarsheechatterjee/VolumeButtonListener/VolumeButtonListener.java
SkillGG: The native module is ONLY to notify ReactNative that the button has been pressed. The rest of the behaviour SHOULD BE IN REACT NATIVE not in java...

PR #598, opened 2023-04-09
bug and theme fix + display option added 

CD-Z:
- fixed colors of midnightdusk since it was the same as Strawberry Daiquiri
- fixed navigationbar visible after closing drawer in readerscreen
- removed several ESLint warnings
- fixed missing key warning in historyscreen
- fixed app crashing after reaching end of chapter (probably because of react native upgrade)

- added removePrefix option in novelscreen bottomsheet to remove chapter in front of the chapternumber


src/screens/novel/components/ChapterItem.tsx
```diff
@@ -56,6 +58,16 @@ const ChapterItem: React.FC<ChapterItemProps> = ({
   const hideDeleteChapterMenu = () => setDeleteChapterMenuVisible(false);
   const chapterNumber = parseChapterNumber(chapterName);
 
+  let processedChapterName = chapterName;
+  if (removePrefix && chapterName) {
+    processedChapterName = chapterName.replace(/[^0-9]*(?=[0-9])/, '');
```
rajarsheechatterjee, line 63: What if the name is `Volume 1 Chapter 8`, wouldn't it show 18?

CD-Z: No, the Regex selects any non number before the first number and since it isn't global, it returns only the first match. So your example would return "1 Chapter 8".

rajarsheechatterjee: Isn't that behaviour incorrect?

CD-Z: You can see it that way, however if we want to also remove the chapter part then it firstly wouldn't look very good in my opinion. I mean you would get "1 8 Chaptername". Secondly we would have to create a big Regex, else it can easily break. For Example with a chaptername such as "Ch. 50 - Level 10" you could potentionally get "50 10". 

My goal with that feature was just to shorten the title a bit, because I enjoy reading the next few titles when opening the drawer and currently they are often cut off because of one word.

rajarsheechatterjee: > My goal with that feature was just to shorten the title a bit, because I enjoy reading the next few titles when opening the drawer and currently they are often cut off because of one word.

This seems like a specific use case, which would end up creating an inconsistent behaviour for others. We could also wrap the chapter names in a [tooltip](https://callstack.github.io/react-native-paper/docs/components/Tooltip/) so that you could long press to view the full text.

CD-Z:
@rajarsheechatterjee 
I changed the remove Prefix option to a tooltip in the drawer.

src/screens/history/components/HistorySkeletonLoading.tsx
```diff
@@ -16,7 +16,7 @@ const HistorySkeletonLoading: React.FC<Props> = ({ theme }) => {
 
   const renderLoadingChapter = (item: number, index: number) => {
     return (
-      <>
```
rajarsheechatterjee, line 19: Index shouldn't be used as a key. You can read more [here](https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern/).

rajarsheechatterjee (approved):

PR #631, opened 2023-05-11
Normalized trackers refactor (+AniList)

### LINKED ISSUE #42: [Feature request] Add Anilist

TehNut:
Closes #42 (Example videos can be found in the replies)

Refactors the tracker system to be normalized for easy support of many different trackers. User data is migrated to the new  data format when they launch the app for the first time since this update.

This was designed mostly with MyAnimeList and AniList in mind, but should be applicable to other trackers as well.

I tested everything I can think of, but I'm not a MAL user (and MAL is dead at the time of opening this PR 🙄) so it could definitely use a once over by another set of eyes. The data migration part could definitely use a solid test.

The `MalTopNovels` and `AniListTopNovels` components can probably be normalized as well, but I'm no React guy so I'm not sure how to "correctly" compose those.

I did not go through and add typing to most of the tracker store, but the new storage model looks like this:
```ts
import type {
  AuthenticationResult,
  SearchResult,
  TrackerName,
  UserListEntry,
} from 'src/services/Trackers';

type Tracker = {
  /** The name of this tracker */
  name: TrackerName;
  /** The authentication data */
  auth: AuthenticationResult;
};

type TrackedNovel = Pick<SearchResult, 'id' | 'totalChapters'> & {
  /** The internal LNReader ID */
  novelId: number;
  /** User specific list data */
  userData: UserListEntry;
};

type TrackerReducer = {
  tracker: Tracker;
  trackedNovels: TrackedNovel[];
};
```


rajarsheechatterjee (approved):

rajarsheechatterjee:
Tested MAL, LGTM.

PR #634, opened 2023-05-13
fix(fetch): Add capacity to change UA string

Rojikku:
# Improves on Fetch.ts
1. Adjusted UA string to my UA string on my Pixel5A running Android 13. User Agents have become more obscured and generic, so having a very specific UA string will start to stand out.
2. Implemented the capacity to specify a unique user agent string, or remove it entirely. Now done by switching line order and adding comments.

Why, you ask, would I want to remove a user agent string? Because if there's no UA string you can get the page without lazy loading images, which can potentially make or break pulling images. Sure, I can just use a raw fetch, but that doesn't have cloudflare bypass. (Specifically for ReaperScans)
This is the best of both worlds, while still functioning as expected with existing scripts.

Theoretically fixes https://github.com/LNReader/lnreader-sources/issues/473 but webview crashes my emulator.
Did several test loads to see if I saw any unexpected errors produced by this conversion, but wasn't able to find any.

Good news: Throws proper errors when cloudflare can't be bypassed instead of giving broke chapters
Bad news: I don't think it's getting the cookie for the bypass.

### DIFF
--- a/src/sources/en/novelupdates.js
+++ b/src/sources/en/novelupdates.js
@@ -1,4 +1,5 @@
 import * as cheerio from 'cheerio';
+import { fetchApi, fetchHtml } from '@utils/fetch/fetch';
 import { defaultTo } from 'lodash-es';
 import { FilterInputs } from '../types/filterTypes';
 const sourceId = 50;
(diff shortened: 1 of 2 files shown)


Rojikku:
Theoretically fixes lnreader/lnreader-sources#473 but webview crashes my emulator.
Did several test loads to see if I saw any errors produced by this conversion, but wasn't able to find any.

rajarsheechatterjee (approved):

Issue #467, opened 2022-11-12
[F-R] Epub Export


Humam-Hamdan:
### Describe your suggested feature

The Option to export the downloads to an epub file, which would make the read on other devices (i.e. kindle) much easier.

### Other details

Like what ranobe has for example

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with a source, I should be opening an issue in the [sources repository](https://github.com/LNReader/lnreader-sources/issues/new/choose).
- [X] I have updated the app to version **[1.1.13](https://github.com/LNReader/lnreader/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

rajarsheechatterjee:
Duplicate of #38

Humam-Hamdan:
@rajarsheechatterjee isn't #38 about "importing" existing epub files and not "exporting what I downloaded"?

rajarsheechatterjee:
> @rajarsheechatterjee isn't https://github.com/LNReader/lnreader/issues/38 about "importing" existing epub files and not "exporting what I downloaded"?

I thought the old ticket was about all kinds of EPUB support. I can reopen this one just to track the export feature.

Humam-Hamdan:
Thanks :), that's what made me open this

> Request local epub support as that would allow readers to access their personal collections.

and

> placed a local file option

PR #646, opened 2023-05-29
Export to Epub

CD-Z:
Adds a possibility to export downloaded Chapters to an EPUB file.

Closes: #467


src/screens/novel/components/EpubIconButton.tsx
```diff
@@ -0,0 +1,156 @@
...
+    color: ${theme.primary};
+  }
+  img {
+    display: block;
+    width: auto;
+    height: auto;
+    max-width: 100%;
+  }
+  
+  .nextButton,
```
rajarsheechatterjee, line 65: If we are directly getting the chapter from DB these styles wouldn't be there, right?

CD-Z: I don't know what exactly you mean.
- First why shouldn't the styles be there
- and second at the moment this feature only takes the already downloaded chapters from the DB and transforms it into an epub

CD-Z: Oh right, I was to lazy to remove the not needed css and later forgot to remove it.

rajarsheechatterjee (approved):

Issue #38, opened 2021-06-20
[Feature Request] Local Epub support


joseph619:
A lot of good translations/official translations are not available on the aggregator sites, and can only be obtained locally, Request local epub support as that would allow readers to access their personal collections.

KasukeLp23:
Yes, it would be greatly appreciated if the creator placed a local file option either in Epub or Pdf format.

BhaturaGuy:
How tf is this not a feature already

oromis995:
> How tf is this not a feature already

It's a volunteer project, have some respect. If you want it so bad why don't you develop it?

cactusoftheday:
My guy we're slowly working on it with our free time. You can also work on it if you'd like. If you know any react native or anything, there's a pull request for this issue but it needs a front end component.

BhaturaGuy:
Sorry, I was drunk last night and I apologise for any disrespect. But from a user perspective this just looks like LNScrapper at best, there isn't a single good source for LNs (ref piracy.moe), there were only MTL and bad fan translations only. Maybe add new sources (good ones) until local EPUB is supported?

PR #649, opened 2023-05-31
epub parser for LNReader

cactusoftheday:
I've left a little bit of code within src/main/java/.../epubParser if you want to know how the file paths are retrieved and passed in. You can probably just copy-paste some of the code to suit your own needs. 

I am a little bit unsure of how the program will run since I was on compileSdkVersion 32 and not 31. MinSdkVersion was 21 across both projects. If the app breaks, please let me know. I might have forgotten some stuff.

Still totally not sure how you will do this with react native but I'm sure you'll find a way.


cactusoftheday:
Accidentally copy pasted requirements.txt to the wrong directory and commented out some floating code. I just did a npm run android and the app was responsive.

cactusoftheday:
Closes #38

cactusoftheday:
https://github.com/LNReader/lnreader/pull/754

PR #657, opened 2023-06-07
fix for sources

K1ngfish3r:
addresses the following issues:
https://github.com/LNReader/lnreader-sources/issues/470
https://github.com/LNReader/lnreader-sources/issues/518
https://github.com/LNReader/lnreader-sources/issues/508
https://github.com/LNReader/lnreader-sources/issues/504
https://github.com/LNReader/lnreader-sources/issues/494
https://github.com/LNReader/lnreader-sources/issues/483
https://github.com/LNReader/lnreader-sources/issues/519 (partial fix)
https://github.com/LNReader/lnreader-sources/issues/360
added lightnovelpub search function
added more check to cloudflarecheck for site that loads halfway
added fastnovel fix


src/sources/multisrc/wqmangastream/WQMangaStreamGenerator.js
```diff
@@ -11,4 +11,5 @@ export const KolNovelScraper = new WQMangaStreamScraper(
   53,
   'https://kolnovel.com/',
   'KolNovel',
+  { reverseChapters: true, ignoredStrings: true },
```
rajarsheechatterjee, line 14: I meant `ignoredStrings` should be an array of strings which would be removed from the chapterText. 

If it is a source specific issue then we should add a check for sourceId instead so it doesn't affect other sources using the multiSrc generator.

K1ngfish3r: An array? The generated text is currently all under the same class value so my solution was to check for the p.(class) within the chapter and remove every instance of it. The (class) value is not fixed, so my implementation is to find out the class value required, and apply accordingly. I'm not quite sure how I'm to make this an array though.

Yes, it is source specific so i can remove it as an option and make it target the sourceId instead. I'm not exactly sure how it would affect other sources within the multiSrc since I'm using it as an `if (this.ignoredStrings)` statement.

K1ngfish3r:
```suggestion
    let chapterName = loadedCheerio('.entry-title').text();
    let chapterText = loadedCheerio('.epcontent').html();

    if (sourceId === 53) {
      let ignore = loadedCheerio('.epcontent > p').next().attr('class');
      loadedCheerio(`p.${ignore}`).remove();
      chapterText = loadedCheerio('.epcontent').html();
    }
```

K1ngfish3r:
please see latest commit, since it targets only one object within the chapterText I did not make it array like in novelupdates.

rajarsheechatterjee (approved):

PR #671, opened 2023-06-25
fix(Source): Arnovel

K1ngfish3r:
to fix https://github.com/LNReader/lnreader-sources/issues/525
points to correct url separator
- add update to source url
1stkissnovel.love -> 1stkissnovel.org
- Fix spacing in Agitoon
- Remove pawread text in chapters

### DIFF
--- a/src/sources/id/indowebnovel.js
+++ b/src/sources/id/indowebnovel.js
@@ -51,6 +51,12 @@ const parseNovelAndChapters = async novelUrl => {
   };
 
   novel.novelName = loadedCheerio('.series-title h2').text().trim();
+  const chapName = novel.novelName
+    .split(' ')
+    .map(chap => {
+      return chap[0].toUpperCase() + chap.slice(1);
+    })
+    .join(' ');
 
   novel.novelCover = loadedCheerio('.series-thumb img').attr('src');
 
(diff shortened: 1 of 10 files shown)


src/sources/id/indowebnovel.js
```diff
@@ -78,22 +84,23 @@ const parseNovelAndChapters = async novelUrl => {
...
-      .first()
-      .next()
-      .text();
-    const chapterUrl = loadedCheerio(this).find('a').attr('href');
+    let titlename = titles
+      .split(' ')
+      .map(title => {
+        return title[0].toUpperCase() + title.slice(1);
+      })
+      .join(' ');
```
rajarsheechatterjee, line 99: You could use `startcase` from lodash if you want to capitalize each word on a string.

src/sources/id/sakuranovel.js
```diff
@@ -46,6 +46,12 @@ const parseNovelAndChapters = async novelUrl => {
   };
 
   novel.novelName = loadedCheerio('.series-title h2').text().trim();
+  const chapName = novel.novelName
+    .split(' ')
+    .map(chap => {
+      return chap[0].toUpperCase() + chap.slice(1);
+    })
+    .join(' ');
```
rajarsheechatterjee, line 54: Same as above.

src/sources/id/sakuranovel.js
```diff
@@ -73,22 +79,23 @@ const parseNovelAndChapters = async novelUrl => {
...
-      .first()
-      .next()
-      .text();
-    const chapterUrl = loadedCheerio(this).find('a').attr('href');
+    let titlename = titles
+      .split(' ')
+      .map(title => {
+        return title[0].toUpperCase() + title.slice(1);
+      })
+      .join(' ');
```
rajarsheechatterjee, line 94: Same as above.

src/sources/multisrc/madara/MadaraScraper.js
```diff
@@ -79,11 +87,12 @@ class MadaraScraper {
     novel.novelName = loadedCheerio('.post-title h1').text().trim();
 
     novel.novelCover =
+      loadedCheerio('.summary_image > a > img').attr('data-lazy-src') ||
       loadedCheerio('.summary_image > a > img').attr('data-src') ||
       loadedCheerio('.summary_image > a > img').attr('src') ||
       defaultCoverUri;
 
-    loadedCheerio('.post-content_item', '.post-content').each(function () {
+    loadedCheerio('.post-content_item' || '.post-content').each(function () {
```
rajarsheechatterjee, line 95: Wouldn't `'.post-content_item'` always be true and it'd never go to the second condition. 

Also is the previous code working as I think loadedCheerio takes a selector as input?

```ts
loadedCheerio('.post-content_item', '.post-content').each(function () {

src/sources/id/indowebnovel.js
K1ngfish3r: Lodash not available in plugins repo soooo...

Idk how to add 😇

src/sources/id/sakuranovel.js
K1ngfish3r: Ditto

src/sources/id/sakuranovel.js
K1ngfish3r: Ditto

src/sources/multisrc/madara/MadaraScraper.js
K1ngfish3r: I figured out how commas are supposed to work. Now it properly obtains the union of the two selectors as defined here https://webplatform.github.io/docs/css/selectors/

src/sources/id/indowebnovel.js
rajarsheechatterjee: It's available in this branch and it'd be easier to maintain.

In plugins, you could create a util function for it instead of repeating.

K1ngfish3r: i added a startCase
i think

rajarsheechatterjee (approved):

PR #710, opened 2023-07-27
refactor(source): Use fetchHtml for Linovelib

alistair3149:
Should also fix the `fetchApi` unrefined error

### DIFF
--- a/src/sources/ch/linovelib.js
+++ b/src/sources/ch/linovelib.js
@@ -81,7 +81,7 @@ const parseNovelAndChapters = async novelUrl => {
   const novelId = url.match(idPattern)[1];
 
   const chaptersUrl = baseUrl + loadedCheerio('#btnReadBook').attr('href');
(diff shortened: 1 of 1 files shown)


src/sources/ch/linovelib.js
```diff
@@ -96,8 +81,7 @@ const parseNovelAndChapters = async novelUrl => {
   const novelId = url.match(idPattern)[1];
 
   const chaptersUrl = baseUrl + loadedCheerio('#btnReadBook').attr('href');
-  const chaptersResult = await fetchApi(chaptersUrl);
-  const chaptersBody = await chaptersResult.text();
+  const chaptersBody = await fetchHtml({ chaptersUrl, sourceId });
```
rajarsheechatterjee, line 84:
```suggestion
  const chaptersBody = await fetchHtml({ url: chaptersUrl, sourceId });
```

PR #715, opened 2023-07-29
Fix: Update Kolnovel & Agitoon

K1ngfish3r:
Updates for Kolnovel chapter scraping and Agitoon popularNovels screen

### DIFF
--- a/src/sources/multisrc/wqmangastream/WQMangaStreamScraper.js
+++ b/src/sources/multisrc/wqmangastream/WQMangaStreamScraper.js
@@ -128,8 +128,12 @@ class WQMangaStreamScraper {
     let chapterText = loadedCheerio('.epcontent').html();
 
     if (sourceId === 53) {
-      let ignore = loadedCheerio('.epcontent > p').next().attr('class');
-      loadedCheerio(`p.${ignore}`).remove();
+      let ignore = loadedCheerio('#post-27775 > style')
+        .text()
+        .trim()
+        .split(',');
+      ignore.push(...ignore.pop().match(/^\.\w+/));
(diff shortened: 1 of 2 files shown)


src/sources/multisrc/wqmangastream/WQMangaStreamScraper.js
```diff
@@ -128,8 +128,9 @@ class WQMangaStreamScraper {
     let chapterText = loadedCheerio('.epcontent').html();
 
     if (sourceId === 53) {
-      let ignore = loadedCheerio('.epcontent > p').next().attr('class');
-      loadedCheerio(`p.${ignore}`).remove();
+      let ignore = loadedCheerio('article > style').text().trim().split(',');
+      ignore.push(...ignore.pop().match(/^\.\w+/));
```
rajarsheechatterjee, line 132:
```suggestion
      let ignore = loadedCheerio('article > style').text().trim().split(',').pop().match(/^\.\w+/);
```

K1ngfish3r: line 131 obtains an array of items from the 'article > style' selector.
The last element of the array contains text which are unneeded for line 133.
So I pop it, match the required pieces and push it back in

The suggested solution only obtains one item (ie: the last item), which is not enough to clean the text

rajarsheechatterjee (approved):

Mingo3347:
How do I get the fixed version of kolnovel

rajarsheechatterjee:
> How do I get the fixed version of kolnovel

You could download a dev build from [here](https://github.com/LNReader/lnreader/suites/14788417616/artifacts/841397047).

PR #739, opened 2023-08-17
Update MadaraGenerator.ts

yemenihandsome:


rajarsheechatterjee:
You would also have to change the URL [here](https://github.com/LNReader/lnreader/blob/main/src/sources/sources.json#L627).

PR #754, opened 2023-08-22
Epub parser for LNReader

cactusoftheday:
Literally just my branch off of nyagami's. GL to whoever is working on this.


rajarsheechatterjee (approved):
Thanks @cactusoftheday and @danisty for working on this.


Few points:

1. The app's style options wouldn't override EPUB's styles.
2. We can have a Local Source option in sources tab which shows all the EPUBs present in a directory set by the user.

danisty:
Thank you for taking the time to test it.

1. The EPUB's styles are loaded as a `style` element, I'd have to look into how the app changes styles.
2. That's a good idea! I'm down to implementing it as soon as I have some free time.

PR #832, opened 2023-11-05
Updated NovelOnline.ts

NewbieAnimeManga:
https://github.com/LNReader/lnreader-sources/issues/638


NewbieAnimeManga:
What is the problem with eslint

PR #888, opened 2024-01-09
remove mmkv cookie storage, use native cookie storage (cloudflare)

nguyd1:
1. fetch on react native is made by the native networking stack so the cookie is stored in the device itself, no need for mmkv storage: https://stackoverflow.com/questions/41132167/react-native-fetch-cookie-persist

2. android syncs the cookies every 5 min, if the app is killed during that time, the cookies will be lost and cloudflare will trigger again, we can fix this by using CookieManager.flush() to force cookie sync when the nav state changes on webview (occurs when cloudflare passes and navigates to website)

3. the old implementation in useEffect was not storing any cookies because it is called right when webview opens and tries to obtain cookies before cloudflare ran, we needed to wait 5-10 seconds for the cloudflare pass to obtain the cf_clearance cookie and write to mmkv source storage

### DIFF
--- a/src/screens/WebviewScreen/WebviewScreen.tsx
+++ b/src/screens/WebviewScreen/WebviewScreen.tsx
@@ -1,11 +1,8 @@
-import React, { useEffect } from 'react';
+import React from 'react';
 import { RouteProp, useNavigation, useRoute } from '@react-navigation/native';
 import WebView from 'react-native-webview';
-import CookieManager from '@react-native-cookies/cookies';
-
 import { Appbar } from '@components';
 import { useTheme } from '@hooks/useTheme';
-import useSourceStorage from '@hooks/useSourceStorage';
 import { defaultUserAgentString } from '@utils/fetch/fetch';
 
 type ReaderScreenRouteProps = RouteProp<{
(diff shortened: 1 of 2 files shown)


nyagami:
Plus, If Android couldstore the cookies it self and cookies are persisted (as the link), users just need to open source web in Chrome or any browser, then back to the app and they should have cookies stored in device, no need to open webview in app.
But in fact that it didnt work .

nguyd1:
> It's not true, you still have to provide headers to make network request. You can test it by making a simple http server then request to it by React Native.

yes the fetch is still sending the user-agent headers, it is also sending the cookies via native storage, i tested the changes on my local and it is working but u can take a look if i am missing something

when i was debugging on main earlier i was noticing the cookies were not added to the headers from the mmkv storage but still passed after webview cloudflare authentication, leading me to find the points above:
image
image

nguyd1:
> Plus, If Android couldstore the cookies it self and cookies are persisted (as the link), users just need to open source web in Chrome or any browser, then back to the app and they should have cookies stored in device, no need to open webview in app. But in fact that it didnt work .

i think this is possible if we use sharedCookiesEnabled on the webview component but ill have to test it out

nguyd1:
https://github.com/facebook/react-native/commit/274c5c78c4933a09335fca0e2fdc266c80fe4017

this is the commit that adds a persistent cookie storage and shares cookies between the webview and the react app

rajarsheechatterjee:
We are also sending the cookie from MMKV in FastImage's headers. We should also remove them if they're working as expected. 

Also thanks for you recent contributions. 🎉

nguyd1:
https://github.com/LNReader/lnreader/pull/896 new pr i accidentally closed this and cant reopen it

PR #914, opened 2024-01-26
Fix: pre-alpha-patch-3

nyagami:
- Not allow users to delete local category
- Remove category which has same id or sort when restoring
- Github cache issue
- english key for plugins


src/hooks/persisted/usePlugins.ts
```diff
@@ -156,6 +160,9 @@ export default function usePlugins() {
 
   const updatePlugin = (plugin: PluginItem) => {
     return _update(plugin).then(_plg => {
+      if (plugin.version === _plg?.version) {
+        throw new Error('Try again in a moment.');
```
he0119, line 164: Maybe we could translate this error message and stop accumulating hard coded strings.

nyagami: the configurations are too complex for quick usage :(

he0119: I have provided the configuration in my pr.

PR #968, opened 2024-02-23
Upgrade TTS 

nyagami:


nyagami:
https://github.com/LNReader/lnreader/assets/86464880/04a11d77-b524-41e8-9c90-9b11716a6c1a

now users can select where to start TTS

PR #975, opened 2024-02-25
webview improved

Rider21:
WebView:
 - added arrow navigation
 - added `refresh` button
 - added `share` button
 - added `open in browser` button
 - added clear cookies and cache button

plugins:
 - added `qs` module
 - added ~`expandURL`~ `resolveUrl` function (to shorten `path` in books, chapters)


src/screens/WebviewScreen/components/Appbar.tsx
```diff
@@ -0,0 +1,101 @@
...
+          onPress={() => WebBrowser.openBrowserAsync(currentUrl)}
+        />
+        <Menu.Item
+          title={getString('webview.clearCookies')}
+          titleStyle={{ color: theme.onSurface }}
+          onPress={() => {
+            CookieManager.clearAll();
+            showToast(getString('webview.cookiesCleared'));
+          }}
+        />
```
nyagami, line 87: I think this button can annoy users if missclick. Users have to bypass all cf plugins again

Rider21: Oops, I think I forgot to delete it.
I moved it to advanced settings.

src/services/plugin/fetch.ts
```diff
@@ -57,3 +58,22 @@ export const fetchPage = async (
   });
   return res;
 };
+
+export const expandURL = (pluginId: string, isNovel: boolean, slug: string) => {
```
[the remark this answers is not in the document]
Rider21: Should this function be made asynchronous?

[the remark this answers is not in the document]
nyagami: No

src/screens/reader/components/ReaderFooter.tsx
```diff
@@ -68,7 +71,13 @@ const ChapterFooter = ({
...
+          onPress={() =>
+            navigate('WebviewScreen', {
+              name: `${chapter.name} | ${novel.name}`,
+              url: chapter.path,
+              pluginId: novel.pluginId,
+            })
+          }
         >
           <IconButton icon="earth" size={26} iconColor={theme.onSurface} />
         </Pressable>
```
nyagami, line 83: Could you add condition `novel.pluginId !== 2`. Because local plugin is not an actual instance

src/screens/reader/components/ReaderFooter.tsx
```diff
@@ -37,6 +39,7 @@ const ChapterFooter = ({
   };
 
   const insets = useSafeAreaInsets();
+  const { navigate } = useNavigation();
```
nyagami, line 42:
```suggestion
  const { navigate } = useNavigation<NavigationProp<RootStackParamList>>();
```

src/screens/WebviewScreen/WebviewScreen.tsx
```diff
@@ -1,35 +1,60 @@
...
 import { WebviewScreenProps } from '@navigators/types';
 import { getUserAgent } from '@hooks/persisted/useUserAgent';
-import { isUrlAbsolute } from '@plugins/helpers/isAbsoluteUrl';
-import { getPlugin } from '@plugins/pluginManager';
+import { resolveUrl } from '@services/plugin/fetch';
+import Appbar from './components/Appbar';
 
 const WebviewScreen = ({ route, navigation }: WebviewScreenProps) => {
+  const { name, url, pluginId, isNovel } = route.params;
+  const uri = pluginId ? resolveUrl(pluginId, url, isNovel) : url;
```
nyagami, line 13: because this is potentially error, I think try catch is needed.
```ts
const [uri, setUri] = useState(url);

useEffect(() => {
 try{
  // do smt
 } catch(e) {
  // do smt
 }
},[]);
```

Rider21: It might be better to return a try catch construct in the resolveUrl function

nyagami: yea

src/services/plugin/fetch.ts
```diff
@@ -57,3 +58,25 @@ export const fetchPage = async (
...
+  const plugin = getPlugin(pluginId);
+  try {
+    if (!plugin) {
+      throw new Error(`Unknown plugin: ${pluginId}`);
+    }
+    if (plugin.resolveUrl) {
+      return plugin.resolveUrl(path, isNovel);
+    }
+  } catch (e) {
+    throw e;
```
nyagami, line 79:
```suggestion
    return path;
```
app will crash if plugin is not installed or plugin.resolve doesnt work

src/screens/reader/components/ReaderFooter.tsx
Rider21: I added hiding the icon with `islocal`.

nyagami (approved):

PR #986, opened 2024-03-04
Added database for plugins

Rider21:
Added database, ~~cookie manager~~ for plugins
Added copying from webview [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localstorage), [sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)


nyagami:
isnt cookie managed by React Native Network?

Rider21:
Yes they are managed, but you can't interact with them, I just made a `react-native-cookies/cookies` wrapper so that from the plugin you can add, modify, check, delete, clear cookies

nyagami:
> Yes, they are managed, but you can't interact with them, I made a wrapper under react-native-cookies/cookies, so from the plugin you can add, modify, check, delete site cookies

But is cookie management that important. Instead of remove/edit one cookie, users can just open webview, old cookies will be overrided.

Rider21:
I don't know, but this process can be automated with these functions.

Rider21:
By the way, do you think I need to redo copying from `webview` of `localStorage`, `essionStorage` objects via java (directly).
Right now it is implemented through injecting `js` code

nyagami:
> I don't know, but this process can be automated with these functions.

Cookies have been already autmatically added by React Native. So I think cookies management isnt neccessary at all.

Rider21:
I think it would be handy to have that option.
example
What if [source](https://github.com/LNReader/lnreader-sources/blob/2061abbd10e2153c11b2d4ba42e477547ced9fe6/plugins/russian/authortoday.ts) has 2 APIs

1. with authorization only
2. without authorization
How to know if the user is authorized on the site?

Or another example, what if the site for some reason will store in cookies what is needed for queries?

nyagami:
- for the authorization issue, I think you just force users to login via webview like how people are using Novel Updates. 
- there's no way a website would decode cookie in client site. security risks

Rider21:
> * there's no way a website would decode cookie in client site. security risks

If I'm not mistaken, there is a `document. cookie`.
Only it stores the cookie for the current page `name=value; `.
So the client has access to at least read the cookie.

src/plugins/helpers/storage.ts
```diff
@@ -0,0 +1,97 @@
...
+    key: string,
+    value: any,
+    expires?: Date | number,
+  ): void {
+    const item: StoredItem = {
+      created: new Date(),
+      value,
+      expires: expires instanceof Date ? expires.getTime() : expires,
+    };
+    this.mmkv.set(`${pluginID}_DB_${key}`, JSON.stringify(item));
```
nyagami, line 29: I think you should use a prefix like `PLUGIN_STORAGE`

src/plugins/pluginManager.ts
```diff
@@ -119,6 +121,11 @@ const installPlugin = async (url: string): Promise<Plugin | undefined> => {
 
 const uninstallPlugin = async (_plugin: PluginItem) => {
   plugins[_plugin.id] = undefined;
+  storage.mmkv.getAllKeys().forEach(key => {
+    if (key.startsWith(_plugin.id)) {
+      storage.mmkv.delete(key);
```
nyagami, line 126: why do you use mmkv's delete method instead of storage instance's one

Rider21: It doesn't have access to `localStorage`, `sessionStorage`.

src/plugins/helpers/storage.ts
```diff
@@ -0,0 +1,120 @@
...
+    key: string,
+    value: any,
+    expires?: Date | number,
+  ): void {
+    const item: StoredItem = {
+      created: new Date(),
+      value,
+      expires: expires instanceof Date ? expires.getTime() : expires,
+    };
+    store.set(pluginID + PLUGIN_STORAGE + key, JSON.stringify(item));
```
nyagami, line 35: for set and get json object, you can reuse the  `getMMKVObject<T>` and `setMMKVObject<T>` in utils/mmkv

src/plugins/pluginManager.ts
nyagami:
```js
  clearAll(pluginID: string): void {
    const keysToRemove = this.getAllKeys(pluginID);
    keysToRemove.forEach(key => this.delete(pluginID, key));
  }
  ```
  but you declare this

File: src/plugins/helpers/storage.ts (hunk 3 above)
Comment (Rider21) [reply src/plugins/helpers/storage.ts:35]: I am using a different instance of `mmkv`.
```ts
const store = new MMKV({ id: 'plugin_db' });
```

The default mmkv instance is used in `utils/mmkv`.
```ts
export const MMKVStorage = new MMKV();
```

src/plugins/pluginManager.ts
Rider21: The `storage.getAllKeys` function will only return keys that were set using `storage.set`.
It will not return keys to `localStorage`, `sessionStorage`.
(these can only be retrieved via a separate instance and are read-only, [example](https://github.com/Rider21/lnreader-sources/blob/db_2/plugins/russian/ranobelib.ts)).

Rider21: I think I forgot to change the description of the methods.

src/plugins/helpers/storage.ts
nyagami: I think we should use the default MMKV instance as singleton instead creating a new one. This is also recommended by the author of react-native-mmkv

src/plugins/helpers/storage.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+   * Gets the value associated with a key from the storage.
+   *
+   * @param pluginID - The ID of the plugin.
+   * @param key - The key to retrieve.
+   * @param raw - Optional. If true, returns the raw storage item object.
+   * @returns The value associated with the key or undefined if not found or expired.
+   */
+  get(pluginID: string, key: string, raw?: boolean): any {
```
nyagami, line 46:
```suggestion
  get(pluginID: string, key: string, raw?: boolean): StoredItem | undefined {
```

nyagami:
There're some vulnerabilities,I can set/get/delete stored values in plugin A from plugin B code right?
Since we had plugin repositories feature, users could be nuked by some trollers

Rider21:
Yes, it's true, it can be fixed. 
But I don't see the point of fixing it. The running plugin has too many permissions.
They could just write something like.
```ts
import { MMKV } from 'react-native-mmkv';

const store = new MMKV({ id: 'plugin_db' });
store.clearAll();
```

nyagami:
> Yes, it's true, it can be fixed. But I don't see the point of fixing it. The running plugin has too many permissions. They could just write something like.
> 
> ```ts
> import { MMKV } from 'react-native-mmkv';
> 
> const store = new MMKV({ id: 'plugin_db' });
> store.clearAll();
> ```

No, plugin does not have that privilege. They are literally standalone scripts and dont have module accesses.
The `import` keyword in plugin code is just a hack. `import` would be converted into `declared require` function which is refer to https://github.com/LNReader/lnreader/blob/e2d22a5b05a47b57e1f46ccfb6d8dd3ebb05bcc8/src/plugins/pluginManager.ts#L23C1-L38C3

nyagami:
LGTM, Im going to test in devices

nyagami:
You need to resolve Typescript errors. My editor is screaming

nyagami (approved):

PR #998, opened 2024-03-10
Feature: dynamic novel appbar

CD-Z:
+ added a dynamic appbar that fades in the backgroundcolor dependent on scroll position
+ updated deps with non breaking changes


nyagami:
I think the appbar should have color like Mihon app (primaryContainer ig). Currently it has same color as the background
Screenshot_20240317-233008

nyagami (approved):

PR #1008, opened 2024-03-20
Update indowebnovel.js

AhmadNaruto:


K1ngfish3r:
Welcome! We have since moved all our source/plugin development over to the LNReader/lnreader-sources repo for LNReader 2.0. You may commit this PR over to the`master` branch there, it would be greatly appreciated.

PR #1021, opened 2024-04-03
Page Reader

CD-Z:
Added an improved version of my Page Reader PR in the main branch. #733 

Improvements:
- pages instead of percent
- working scrollbars
- Chapter switch at the end or beginning of a chapter like a normal page
- autoscroll to last position

Additional changes:
- added change padding setting to reader
- fixed scroll to top button in reader


src/screens/reader/components/ReaderBottomSheet/ReaderValueChange.tsx
```diff
@@ -3,43 +3,61 @@ import React from 'react';
...
-interface ReaderLineHeightProps {
+type ValueKey<T extends object> = {
+  [K in keyof T]: T[K] extends number ? K : never;
+}[keyof T];
+
+interface ReaderValueChangeProps {
   labelStyle?: TextStyle | TextStyle[];
+  valueChange?: number;
+  label: string;
+  valueKey: ValueKey<ReturnType<typeof useChapterReaderSettings>>;
```
nyagami, line 15:
```suggestion
  valueKey: ValueKey<ChapterReaderSettings>;
```
for less scary, remember to import `ChapterReaderSettings`

src/screens/reader/components/ReaderBottomSheet/ReaderValueChange.tsx
```diff
@@ -3,43 +3,61 @@ import React from 'react';
...
         <IconButtonV2
           name="minus"
           color={theme.primary}
           size={26}
-          disabled={lineHeight <= 1.3}
+          disabled={settings[valueKey] <= min}
           onPress={() =>
-            setChapterReaderSettings({ lineHeight: lineHeight - 0.1 })
+            setChapterReaderSettings({
+              [valueKey]: settings[valueKey] - valueChange,
```
nyagami, line 44:
```suggestion
              [valueKey]: Math.max(min, settings[valueKey] - valueChange),
```

src/screens/reader/components/ReaderBottomSheet/ReaderValueChange.tsx
```diff
@@ -3,43 +3,61 @@ import React from 'react';
...
         <IconButtonV2
           name="plus"
           color={theme.primary}
           size={26}
-          disabled={lineHeight >= 2}
+          disabled={settings[valueKey] >= max}
           onPress={() =>
-            setChapterReaderSettings({ lineHeight: lineHeight + 0.1 })
+            setChapterReaderSettings({
+              [valueKey]: settings[valueKey] + valueChange,
```
nyagami, line 59:
```suggestion
              [valueKey]: Math.min(max, settings[valueKey] + valueChange),
```

src/screens/settings/SettingsReaderScreen/SettingsReaderScreen.tsx
```diff
@@ -176,14 +185,18 @@ const SettingsReaderScreen = () => {
...
-                  onclick="reader.post({type:'hide'})"
-                >
-                  ${dummyHTML}
-                </chapter>
+              <body> 
+                <div class="chapterCtn"> 
+                  <chapter 
+                    data-page=0
+                    data-pages=0
+                    data-page-reader='${pageReader}'
```
nyagami, line 193: what are they for

src/screens/reader/components/ReaderBottomSheet/ReaderValueChange.tsx
CD-Z: Makes sense, I only thought of checking for a type with the hook declaration.

src/screens/reader/components/ReaderBottomSheet/ReaderValueChange.tsx
CD-Z: I mean we can do it like this, but the button already gets disabled ```ts disabled={settings[valueKey] <= min} ```

src/screens/reader/components/ReaderBottomSheet/ReaderValueChange.tsx
CD-Z: same

src/screens/reader/components/WebViewReader.tsx
```diff
@@ -124,9 +127,39 @@ const WebViewReader: FC<WebViewReaderProps> = props => {
...
+    } catch (e) {}
+
+    if (dataPayload) {
+      if (dataPayload.type === 'console') {
+        // eslint-disable-next-line no-console
+        console.info(`[Console] ${JSON.stringify(dataPayload.msg, null, 2)}`);
+      } else if (false) {
+        // eslint-disable-next-line no-console
+        if (dataPayload.type !== 'save') console.log(dataPayload);
+      }
```
nyagami, line 156: tbh logging via React Native for Webview is not an effecient way. Class objects wouldnt be logged.
If someone want to work in Reader, they can create a html with fake data and all reader js/css then open it in browser, which is eaiser and faster way.

nyagami:
btw I recommend updating pageReader setting without re-rendering by using Reader::updateGengeralSettings (if possilbe)

Any changes in settings are listened by this code (WebviewReader.tsx), then it triggers to Webview without update state in React 
```ts
    const mmkvListener = MMKVStorage.addOnValueChangedListener(key => {
      switch (key) {
        case CHAPTER_READER_SETTINGS:
          webViewRef.current?.injectJavaScript(
            `reader.updateReaderSettings(${MMKVStorage.getString(
              CHAPTER_READER_SETTINGS,
            )})`,
          );
          break;
        case CHAPTER_GENERAL_SETTINGS:
          webViewRef.current?.injectJavaScript(
            `reader.updateGeneralSettings(${MMKVStorage.getString(
              CHAPTER_GENERAL_SETTINGS,
            )})`,
          );
          break;
      }
  ```

src/screens/reader/components/WebViewReader.tsx
CD-Z: Of course this isn't a good way. I just implemented this because of things that behave a bit different in webview, than in the browser. E. g. On opening a new chapter, the chapter width isn't always the same at first. Probably just because of slower calculations but nevertheless a bug I haven't encountered on my browser.

CD-Z:
> btw I recommend updating pageReader setting without re-rendering by using Reader.updateGengeralSettings (if possilbe)
> 
> Any changes in settings are listened by this code (WebviewReader.tsx), then it triggers to Webview without update state in React
> 
> ```ts
>     const mmkvListener = MMKVStorage.addOnValueChangedListener(key => {
>       switch (key) {
>         case CHAPTER_READER_SETTINGS:
>           webViewRef.current?.injectJavaScript(
>             `reader.updateReaderSettings(${MMKVStorage.getString(
>               CHAPTER_READER_SETTINGS,
>             )})`,
>           );
>           break;
>         case CHAPTER_GENERAL_SETTINGS:
>           webViewRef.current?.injectJavaScript(
>             `reader.updateGeneralSettings(${MMKVStorage.getString(
>               CHAPTER_GENERAL_SETTINGS,
>             )})`,
>           );
>           break;
>       }
> ```

I took a look at it and if I understand you correctly you suggest to not rerender the page when changing from the normal view to the page view.

CD-Z:
@nyagami could you please take another look at this, I don't want to always solve the merge conflicts. It gets annoying

nyagami:
Okay

nyagami (approved):
LGTM

rajarsheechatterjee (approved):

PR #1064, opened 2024-05-04
Feature/tailing skeletons

CD-Z:
+ added tailing skeleton novel to complete a row while scrolling in a plugin
+ changed flatlist onEndReachedThreshold to start loading before skeleton novels are visible

I closed the old PR since it was already partly adapted through other PRs .

 
Todo:
- [x] cleanup

### DIFF
--- a/src/components/NovelCover.tsx
+++ b/src/components/NovelCover.tsx
@@ -36,7 +36,34 @@ interface DownloadBadgeProps {
   theme: ThemeColors;
 }
 
-function NovelCover<TNovel extends NovelItem | NovelInfo | LibraryNovelInfo>({
+type coverItemLibrary =
+  | LibraryNovelInfo & {
+      completeRow?: number;
+    };
+
+type coverItemPlugin =
+  | (NovelInfo | NovelItem) & {
+      completeRow?: number;
+    };
+
+function NovelCover<TNovel extends coverItemPlugin>({
+  item,
+  onPress,
+  libraryStatus,
+  theme,
+  isSelected,
+  onLongPress,
+  selectedNovelIds,
+}: {
+  item: TNovel;
+  onPress: () => void;
+  libraryStatus: boolean;
+  theme: ThemeColors;
+  isSelected: boolean;
+  onLongPress: (item: TNovel) => void;
+  selectedNovelIds: number[];
+}): React.ReactNode;
+function NovelCover<TNovel extends coverItemLibrary>({
   item,
   onPress,
   libraryStatus,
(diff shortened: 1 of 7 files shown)


src/components/NovelCover.tsx
```diff
@@ -35,23 +36,37 @@ interface DownloadBadgeProps {
...
-}: {
+type coverItemLibrary =
+  | LibraryNovelInfo & {
+      completeRow?: number;
+    };
+
+type coverItemPlugin =
+  | NovelItem & {
+      completeRow?: number;
+    };
```
nyagami, line 47:
```suggestion
type CoverItemLibrary =
  | LibraryNovelInfo & {
      completeRow?: number;
    };

type CoverItemPlugin =
  | NovelItem & {
      completeRow?: number;
    };
```
just for code conventions

src/screens/BrowseSourceScreen/BrowseSourceScreen.tsx
```diff
@@ -131,6 +131,10 @@ const BrowseSourceScreen = ({ route, navigation }: BrowseSourceScreenProps) => {
                 libraryStatus={inLibrary}
                 onPress={() => navigateToNovel(item)}
                 isSelected={false}
+                addSkeletonLoading={
+                  (hasNextPage && !searchText) ||
+                  (hasNextSearchPage && !!searchText)
```
nyagami, line 136:
```suggestion
                  (hasNextSearchPage && searchText)
```

CD-Z: It would work as how you suggest, but also throw a ts error (can't assign string to boolean). An other option would be to use Boolean().

nyagami: oh ye, I forgot it. But I think using Boolean is better. It is a bit longer but more readable I think

src/screens/browse/loadingAnimation/SourceScreenSkeletonLoading.tsx
```diff
@@ -31,20 +35,22 @@ const SourceScreenSkeletonLoading: React.FC<Props> = ({ theme }) => {
...
+      <View style={{ flex: 1 / numColumns }}>
+        <LoadingNovel
+          key={index}
+          backgroundColor={backgroundColor}
+          highlightColor={highlightColor}
+          pictureHeight={pictureHeight}
+          pictureWidth={pictureWidth}
+          displayMode={displayMode}
+        />
+      </View>
```
nyagami, line 53: use should use unique key for this instead of index.

nyagami (approved):

PR #1072, opened 2024-05-17
About page Version update & GH action update

error7404:
GitHub action: bd2a360 
 - custom apk:
   - packet name with git commit id
   - app name with git commit id
 - cached `npm install`
 - update actions versions
image
image

---
About page: 1635725
 - Version based on package.json
 - added git commit hash to the version if not Stable
 - RELEASE_DATE, GIT_HASH, BUILD_TYPE created at build time from setEnvFile.cjs
 - special BUILD_TYPE for github actions
 - Version modal copy on click
 - removed versionUtils
image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,34 +9,57 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Clone Repository (Latest)
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - uses: actions/setup-node@v3
+      - uses: actions/setup-node@v4
         with:
-          node-version: '18'
+          node-version: '20'
+
+      - name: Cache npm & node modules
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.npm
+            node_modules
+          key: ${{ runner.os }}-cache-npm-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-cache-npm-
+    
       - name: Install Dependencies
         run: |
           npm install
+
       - name: Create .env
         run: |
           cat > .env << EOF
           MYANIMELIST_CLIENT_ID=${{ vars.MYANIMELIST_CLIENT_ID }}
           ANILIST_CLIENT_ID=${{ vars.ANILIST_CLIENT_ID }}
           EOF
+
       - name: Cache Gradle Wrapper
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: ~/.gradle/wrapper
           key: ${{ runner.os }}-gradle-wrapper-${{ hashFiles('gradle/wrapper/gradle-wrapper.properties') }}
 
       - name: Cache Gradle Dependencies
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: ~/.gradle/caches
           key: ${{ runner.os }}-gradle-caches-${{ hashFiles('gradle/wrapper/gradle-wrapper.properties') }}
           restore-keys: |
             ${{ runner.os }}-gradle-caches-
+
+      - name: Cache build intermediates
+        uses: actions/cache@v4
+        with:
+          path: |
+            android/app/build/generated
+            android/app/build/intermediates
+            android/app/build/tmp
+            android/.gradle
+          key: ${{ runner.os }}-build-${{ hashFiles('package-lock.json') }}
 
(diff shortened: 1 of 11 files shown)


rajarsheechatterjee (approved):

package.json
```diff
@@ -3,18 +3,20 @@
...
     "prettier": "prettier --write \"./src/**/*.{js,jsx,ts,tsx}\"",
-    "buildRelease": "cd android && ./gradlew clean && ./gradlew assembleRelease",
+    "buildRelease": "npm run prestart && cd android && ./gradlew clean && ./gradlew assembleRelease",
     "open": "open ./android/app/build/outputs/apk/release/",
     "strings": "node scripts/stringTypes.cjs",
     "prepare": "husky install"
   },
   "dependencies": {
     "@gorhom/bottom-sheet": "^4.4.5",
+    "@react-native-community/clipboard": "^1.5.1",
```
nyagami, line 19: We already had expo-clipboard

nyagami:
I think release time is enough even for dev

env.d.ts
```diff
@@ -1,4 +1,7 @@
 declare module '@env' {
   export const MYANIMELIST_CLIENT_ID: string;
   export const ANILIST_CLIENT_ID: string;
+  export const GIT_HASH: string;
+  export const RELEASE_DATE: string;
+  export const BUILD_TYPE: string;
```
nyagami, line 6:
```suggestion
  export const BUILD_TYPE: 'Release' | 'Debug';
```
Traitor could forgot, and set BUILD_TYPE = 'release', we would have a weird build name in release app

package.json
```diff
@@ -3,18 +3,20 @@
...
   "scripts": {
-    "android": "react-native run-android --active-arch-only",
+    "android": "npm run prestart && react-native run-android --active-arch-only",
     "ios": "react-native run-ios",
+    "prestart": "node scripts/setEnvFile.cjs Debug",
     "start": "react-native start",
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
     "prettier": "prettier --write \"./src/**/*.{js,jsx,ts,tsx}\"",
-    "buildRelease": "cd android && ./gradlew clean && ./gradlew assembleRelease",
+    "buildRelease": "npm run prestart && cd android && ./gradlew clean && ./gradlew assembleRelease",
```
nyagami, line 12: buildRelease but build an apk with Debug in name.

package.json
error7404: I'm blind...

.github/workflows/build.yml
```diff
@@ -9,49 +9,97 @@ jobs:
...
+          path: android/app/build/outputs/apk/release/LNReader.apk
+
+      - name: Build Custom APK
+        env:
+          COMMIT_COUNT: ${{ env.COMMIT_COUNT }}
+          COMMIT_ID: ${{ env.COMMIT_ID }}
+        run: |
+          sed -i 's/LNReader/LNReader-r${{ env.COMMIT_COUNT }}(${{ env.COMMIT_ID }})/g' android/app/src/main/res/values/strings.xml
+          cd android && ./gradlew assembleRelease --no-daemon -PcustomAppId=com.rajarsheechatterjee.LNReader.commit_${{ env.COMMIT_ID }}
+          mv app/build/outputs/apk/release/app-release.apk app/build/outputs/apk/release/LNReader-r${{ env.COMMIT_COUNT }}\(${{ env.COMMIT_ID }}\)-customID.apk
```
nyagami, line 96: Why we should build app 2times.

error7404: For custom name and package id

nyagami: I think only custom one is enough. no need the release one. 
We dont use github action artifact for release one anyway

error7404: When redirecting people to download the latest build it would be better if it didn't install a new app on there phones, so both are useful or only the non-custom

nyagami: We should not suggest people to override the stable build with dev build.  Dev build is more like for `testing` than `quick fix` imo

error7404: well they have the choice

nyagami: Also, we have to consider free github actions 2000 minutes/month.
This will double execution time. it is wasting and abusing

error7404: first build is 9min second is 1min, and i've increased the number of cached things

error7404: i lied it's 9min when cache miss and 5min on cache hit the second one is always 1min tho

PR #1076, opened 2024-05-22
Reader footer & Navigation bar

nyagami:
- Replace native navigation-bar code with native lib 
- fix fullscreen mode logic
- adjust reader footer bottom position: [android navigation bar height](https://forums.solar2d.com/t/height-of-android-navigation-bar-solved/353073)


src/theme/utils/setBarColor.ts
```diff
@@ -1,11 +1,21 @@
...
 import { ThemeColors } from '@theme/types';
+import * as NavigationBar from 'expo-navigation-bar';
+import Color from 'color';
 
-/**
- * Sets status and navigation bar color.
- *
- * @param theme
- */
+export const setBarColor = (color: ThemeColors | Color) => {
```
rajarsheechatterjee, line 6:
```suggestion
export const setStatusBarColor = (color: ThemeColors | Color) => {
```

PR #1080, opened 2024-05-26
Enhancement: Add more CSS to increase Reader Experience

Batorian:
- adding styles for specific classes in specific plugins to increase reader experience


Batorian:
I have no idea what all these small file changes are. I didn't touch them.

rajarsheechatterjee:
We should move the CSS to plugins instead. Otherwise, any changes would depend on a new app release to be reflected.

Batorian:
> We should move the CSS to plugins instead. Otherwise, any changes would depend on a new app release to be reflected.

We could use this as a temporary solution though until we have plugin specific CSS.

PR #1098, opened 2024-06-04
Add open random entry button

Soopyboo32:
Missed this functionality from Mihon for when i'm indecisive
https://youtu.be/POQg-Du9dS8
note: This is my first time using react or react-native so the code may have issues

Also moves the import epub button into a dropdown with the open random button to avoid having too many buttons in top bar

### DIFF
--- a/src/screens/library/LibraryScreen.tsx
+++ b/src/screens/library/LibraryScreen.tsx
@@ -24,7 +24,7 @@ import {
 } from '@hooks/persisted';
 import { useSearch, useBackHandler, useBoolean } from '@hooks';
 import { getString } from '@strings/translations';
-import { FAB, Menu, Portal } from 'react-native-paper';
+import { FAB, Portal } from 'react-native-paper';
 import {
   markAllChaptersRead,
   markAllChaptersUnread,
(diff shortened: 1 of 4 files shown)


Soopyboo32:
just realised it prob makes more sense to create the menu inside SearchbarV2 and just pass in button name & callbacks, will change

src/screens/library/LibraryScreen.tsx
```diff
@@ -139,6 +140,18 @@ const LibraryScreen = ({ navigation }: LibraryScreenProps) => {
     setFalse: closeSetCategoryModal,
   } = useBoolean();
 
+  function openRandom() {
+    const novels = library[index].novels;
+    const randomNovel = novels[Math.floor(Math.random() * novels.length)];
```
nyagami, line 145: Doesn't this apply for current category only?

Soopyboo32: yes, it also respects current filter applied

for reference, my workflow with it is: open want to read category -> open random -> move to reading category

nyagami (approved):

PR #1101, opened 2024-06-11
Fix: Scroll to current Chapter

CD-Z:
Cleaned ChapterDrawer and fixed the Scroll to current Chapter behaviour.


Palloxin:
Does this fix this too?
https://discord.com/channels/835746409357246465/1178895455237701652/1243675525449252935 (discord link to a message in LNReader server)

CD-Z:
Yes, this issue resulted from memoizing the index of the current chapter. The problem was that it returned 0 at first, because the chapter list hasn't been loaded yet. That is why it always points to the first chapter.

nyagami:
And I think using chapter id to detect the list item position is not correct for all novels. 
There some novel has this order: `id: 455`, `id: 456`, `id: 445`,
you should use `chapter.position`, position per novel page or the index inside chapter drawer list

CD-Z:
@nyagami Why? I use the findIndex function to return the index on the chapter drawer list if the id of the currently opened chapter matches the id of the chapter in the list. This has to work, else the navigation to the chapter would also be broken, since this also depends on the id of the chapter in the list.

nyagami:
I mean this. 
```ts
if (
  listAscending
    ? viewableItems[0].item.id < chapter.id
    : viewableItems[0].item.id > chapter.id
)
```

- `chapter.position` is more reliable. 
- The indexes extracted from the list is the most reliable.
Because Flashlist ViewToken has index, we only need the index in the list of current chapter
```ts
interface ViewToken {
  index: number;
  isViewable: boolean;
  item: string;
  key: string;
  timestamp: number;
}
```
And I dont think navigation would be broken no matter what you are using.

nyagami:
Ops, I forgot the custom page novels. you can try the test Hako plugin. 
so I think extracting the current chapter index from the list is the best way =))

CD-Z:
I tested it now for a bit with a few novels and haven't encountered any problems.

src/screens/reader/components/ChapterDrawer/index.tsx
```diff
@@ -65,129 +68,70 @@ const ChapterDrawer = ({
...
+  const checkViewableItems = ({
+    viewableItems,
+  }: {
+    viewableItems: ViewToken[];
+  }) => {
+    let up = {
+      text: getString('readerScreen.drawer.scrollToTop'),
+      func: () => {
+        listRef.current?.scrollToIndex({ index: 0, animated: true });
+      },
```
nyagami, line 80: I think using index (undefinded if scroll to end)  instead of a callback is better.
```ts
const scroll = useCallback((index?: number) => {
  }, [listRef.current]);
```
Also, moving the ListFooter to this file and call `scroll callback` directly  in button oppress could be cleaner.

CD-Z: I now refactored the file and I think I got what you meant. At least the file is now definitly cleaner.

nyagami (approved):

PR #1103, opened 2024-06-11
fix: Smooth ReaderFooter appearance

CD-Z:
fixed the appearance of the ReaderFooter through workaround


nyagami:
What if user changes the orentiation from `landscape` to `portrait`? Doest it look weird

CD-Z:
> What if user changes the orentiation from `landscape` to `portrait`? Doest it look weird

No why should it. It works exactly the same way as at has until now with the difference that it takes the bottom safe area inset into acount.

nyagami:
Try these steps: 
- open chapter
- hide navbar
- swipe to open drawer
- choose another chapter
-> bottom inset now is 0

CD-Z:
> Try these steps:
> 
>     * open chapter
> 
>     * hide navbar
> 
>     * swipe to open drawer
> 
>     * choose another chapter
>       -> bottom inset now is 0

fixed this issue with initialWindowMetrics, it will now always instantly display the safeAreaInsets

PR #1129, opened 2024-07-07
Improvement: Updated Sanitizer

Batorian:
- fix images being displayed and updates sanitizer


src/screens/reader/utils/sanitizeChapterText.ts
```diff
@@ -2,26 +2,87 @@ import { getString } from '@strings/translations';
...
+  // Generate the transformTags object
+  const transformTags: {
+    [key: string]: (tagName: string, attribs: any) => any;
+  } = {};
+  styledTags.forEach(tag => {
+    transformTags[tag] = createTransformFunction();
+  });
+
   let text = sanitizeHtml(html, {
     allowedTags: sanitizeHtml.defaults.allowedTags.concat([
```
CD-Z, line 59: You are extending the default tags. Most of the tags are already included (i think everything except img), so they are not neccessary.

nyagami (approved):

PR #1142, opened 2024-07-13
Convert Clock to 12HR

Cohinem:
Requested by a discord user. and better overall. Aligning with the clock in the history tab.


CD-Z:
It would probably be better to set an option to switch this, else there will be another user requesting the 24hr version because they don't like 12hr.

rajarsheechatterjee:
I agree with CD-Z's comment. Ideally, there should be an option to change it. Changing it for everyone might result in someone requesting a reversal, as it could affect users who are accustomed to the current behavior.

Cohinem:
Well then. I'll have to close this PR. I'm not too familiar with the project to add that.

Issue #861, opened 2023-12-13
Some epub exports do not have TOC titles


spicedcuttlefish:
### Steps to reproduce

1. Download chapters for following (I can provide other sources/titles if necessary):
    * Source: LightNovelPub
    * Title: Soul Land IV (Douluo Dalu) : Ultimate Fighting
2. Rename the title to exclude ":" (This problem is already reported [here](https://github.com/LNReader/lnreader/issues/696))
3. Export to epub
4. Open epub in any reader and look at the TOC

Here are examples of how TOC looks in different readers:
[Calibre](https://cdn.discordapp.com/attachments/907877548489338900/1184571495142395934/image.png?ex=658c752c&is=657a002c&hm=61cd3ae8cafc46bf017d0e8fd7ffe45c80f45cb4901185685c6b3a17f8d179b8&)
[Moon Reader Android App](https://cdn.discordapp.com/attachments/907877548489338900/1184571619755167794/Screenshot_20231213_140430_Moon_Reader_Pro.jpg?ex=658c754a&is=657a004a&hm=cf7e22b414323ab3b716583cf7597d694e761fe7cc6fb7762ee21f89e9b0e6c7&)

Here's how the chapters look in [lnreader](https://cdn.discordapp.com/attachments/907877548489338900/1184571701296631929/Screenshot_20231213_140454_LNReader.jpg?ex=658c755d&is=657a005d&hm=64490794cc45dc1b88d5390c2e542f5f414cfa20c4199d2a2217246bc0c19e39&)

### Expected behavior

The TOC should have either chapter # or the chapter title (preferrably both)

### Actual behavior

TOC chapter names are blank

### LNReader version

1.1.18

### Android version

Android 14

### Device

Samsung Galaxy S23

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with a source, I should be opening an issue in the [sources repository](https://github.com/lnreader/lnreader-sources/issues/new/choose).
- [X] I have updated the app to version **[1.1.18](https://github.com/lnreader/lnreader/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

rajarsheechatterjee:
Resolved in 4cb1890b62eb57cc75ca0628a11be312c930ddc4

Issue #797, opened 2023-10-09
Problem with specific path choosing of epub export


Palloxin:
### Steps to reproduce

1. open export options
2. choose path to raw internal storage (means not in folders inside internal storage, just internal storage)

### Expected behavior

Able to choose that path.

### Actual behavior

https://github.com/LNReader/lnreader/assets/75091899/428abe52-f1f0-472d-bc80-0e86557bcb83

### LNReader version

1.1.18

### Android version

Android 8 miui 8

### Device

huawei p8 lite 2017

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with a source, I should be opening an issue in the [sources repository](https://github.com/lnreader/lnreader-sources/issues/new/choose).
- [X] I have updated the app to version **[1.1.18](https://github.com/lnreader/lnreader/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

CD-Z:
I'm pretty sure the main problem is that there is a wrong error message, since I think that android doesn't allow apps to directly write into the internal storage, only in deeper folders.

Palloxin:
I just tried it on tachiyomi and it works fine.
The path selection and an actual backup creation(it too requires a path selection)
![IMG_20231014_195618](https://github.com/LNReader/lnreader/assets/75091899/946e3e8a-df4c-4dab-909a-60a4b8f78aeb)

Palloxin:
Still happens on 2.1.2

Issue #777, opened 2023-09-08
Export EPUB: exported chapters not properly reconstructed


Palloxin:
### Steps to reproduce

1. go to Panda Machine translations
2. search Into the Creations and download chapter 1
3. export the novel (so only chap 1) with js included(because i did so and i dont want to test if it appears if i dont include JS)
4. check with an editor the html of chap 1 (see images below)

### Expected behavior

![IMG_20230908_153910](https://github.com/LNReader/lnreader/assets/75091899/376563c5-f982-445d-b7e5-de0ff5288a0b)

### Actual behavior

![IMG_20230908_160117](https://github.com/LNReader/lnreader/assets/75091899/0c78b962-7f3b-430d-ac89-e94e973345fd)

Looking into the html file u can see this
![IMG_20230908_155549](https://github.com/LNReader/lnreader/assets/75091899/8bf14c28-ec91-4a22-a9ec-bfc303846d46)
Whatever there was there in the chapter in lnreader, it is not present in the extracted epub.

### LNReader version

1.1.16

### Android version

Android 8 miui 8

### Device

Huawei p8 lite 2017

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with a source, I should be opening an issue in the [sources repository](https://github.com/lnreader/lnreader-sources/issues/new/choose).
- [X] I have updated the app to version **[1.1.17](https://github.com/lnreader/lnreader/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

KrisadaFantasy:
Not just missing on exported EPUB, but probably in imported as well!

I made the EPUB file myself in Calibre and imported it to the app. Found that LNReader is not rendered <br/> properly so I come here to see if there's an existing issue. At first I just thought that the app has issue on rendering HTML or something similar, but @Palloxin screenshot show the downloaded text rendered <br/> correctly. So the issue is something remove <br/> during import/export EPUB process?

Issue #745, opened 2023-08-18
[Little Flaw]Chapters inside the EPUB are not numbered properly


Palloxin:
### Describe your suggested feature

![IMG_20230818_131455](https://github.com/LNReader/lnreader/assets/75091899/5b3961a3-16e7-4924-93eb-6d87dd4e4fd2)
It is from an epub with chapters from chap 1 to chap 10.
The files should be numbered
01
02
03 
and so on (some novels might have the chapter 00)

### Other details

When the chap number is higher like 450 it appears like this:
![IMG_20230818_132415](https://github.com/LNReader/lnreader/assets/75091899/faa71f76-7132-4601-8626-d2db54d20caa)

these are chapters numbered from 455 to 464

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with a source, I should be opening an issue in the [sources repository](https://github.com/LNReader/lnreader-sources/issues/new/choose).
- [X] I have updated the app to version **[1.1.14](https://github.com/LNReader/lnreader/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

PR #1144, opened 2024-07-14
Better export to epub

CD-Z:
I finally found the time to work a bit on the library and implemented it in LNReader with a few changes for better compatibility.
- changed lineheight and textsize to be rem based instead of px
- changed epub text to be wrapped in div with id chapter instead of a chapter element, since the epub standard doesn't accept custom elements.
- added to id chapter to the readerscreen chapter element.

This PR closes:
- #861 
- #797 
- #777 
- #696 
- #745 
  - works as expected
  - shouldn't be viewed in the first place
  - prevents filename duplications
- #740 
  - not app/library related


nyagami:
Why should we export reader ui to epub?

package.json
```diff
@@ -9,13 +9,15 @@
...
     "open": "open ./android/app/build/outputs/apk/release/",
     "strings": "node scripts/stringTypes.cjs",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "cssToJs": "node scripts/convertCssToJs.cjs"
   },
   "dependencies": {
-    "@cd-z/react-native-epub-creator": "^2.0.4",
+    "@cd-z/epub-constructor": "^3.0.1",
+    "@cd-z/react-native-epub-creator": "^3.0.0",
```
nyagami, line 20: Are they duplicated?

package.json
```diff
@@ -51,6 +53,7 @@
     "react-native-device-info": "^8.4.7",
     "react-native-drawer-layout": "^3.2.2",
     "react-native-error-boundary": "^1.1.16",
+    "react-native-file-access": "^3.1.0",
```
nyagami, line 56: Any usage in app?

scripts/convertCssToJs.cjs
```diff
@@ -0,0 +1,29 @@
...
+const path = require('path');
+
+const cssFilePath = path.resolve(
+  __dirname + '/../android/app/src/main/assets/css/',
+  'index.css',
+);
+const jsFilePath = path.resolve(
+  __dirname + '/../android/app/src/main/assets/css/',
+  'index.js',
+);
```
nyagami, line 11: If this file is created in building time only, it should be ignore from git.

src/screens/novel/components/ExportToEpub/ChooseEpubLocationModal.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useState } from 'react';
 import { StyleSheet, View } from 'react-native';
-import { Modal, TextInput, Text } from 'react-native-paper';
+import { Modal, TextInput, Text } from 'react-native-paper'; //@ts-ignore
```
nyagami, line 3:
```suggestion
import { Modal, TextInput, Text } from 'react-native-paper';
```

src/screens/novel/components/ExportToEpub/ExportEpubModal.tsx
```diff
@@ -8,27 +8,29 @@ import EpubBuilder from '@cd-z/react-native-epub-creator';
 import { ChapterInfo, NovelInfo } from '@database/types';
 
 import { useChapterReaderSettings } from '@hooks/persisted';
-import { useBoolean } from '@hooks/index';
 import { showToast } from '@utils/showToast';
 import { NOVEL_STORAGE } from '@utils/Storages';
 import FileManager from '@native/FileManager';
+import color from 'color';
+//@ts-ignore
+import css from '../../../../../android/app/src/main/assets/css/index';
```
nyagami, line 16: for static files, you should import by using `require` to avoiding TS warning

package.json
CD-Z: I fucked up the dependency so i had to create another version and installed it manually to make sure it works, however it can be removed again since it lt's not neccessary.

package.json
CD-Z: It belongs to my package as dependency, however if you don't install it manually, for some reason it says the package is not linked and can't be used. Which means the export feature is broken.

CD-Z:
> Why should we export reader ui to epub?

Where do I export the reader ui?

nyagami:
> > Why should we export reader ui to epub?
> 
> Where do I export the reader ui?

```js
` + (css as string).replace(/chapter/g, '#chapter')
```

nyagami:
Btw, you could use implemented `FileManager.pickerFolder` to pick a folder with resolved path.

Palloxin:
Bug report:
The numeration of xhtml files is based on the sort order set on the novel entry. Force it as ascending.

Uploading YouCut_20240727_181448669.mp4…

How to reproduce:
-pick an entry, download 10 chapters, set chapter list to descending order, export, check chapter files

Palloxin:
Feature request:
Make chapter file numbering start from chapter 1

The majority of webnovels dont have a chapter before chapter 1.

Why? when I edit epubs it gets confusing. Since most webnovels follow perfectly the numbering without derails, maybe it's  a better thing.

CD-Z:
> Feature request: Make chapter file numbering start from chapter 1

Not planed, since only a minority of users will open the epub

Palloxin:
F-r
Add warning in case the downloaded chapters have hole in between.
chapter 1-8 — downloaded
chapter 9-16 — NOT downloaded
chapter 17-100 — downloaded

It would be nice to know. Not necessary tho.

Palloxin:
Performance issue:
Export speed slower in beta(tested on latest)

beta(2:25):

https://github.com/user-attachments/assets/db111b97-305d-4c40-8c79-b7b35d5fa730

Stable(1:10):

https://github.com/user-attachments/assets/2b7bf3f0-3cff-467e-b223-f71c189e9cc9

Palloxin:
Bug

Export failed due to missing images in a chapter. The chapter is downloaded but somehow the images werent, maybe an app bug or the plugin failed to pick them.

-Novelfull 1.0.0, Cultivation chat group, chapter 841

Solution: override or/and make a window appear that says that these chapter failed for some reason

===
EDIT: Also happens when an image is referenced in a html file AND is missing.

Example: I imported an epub. The import process deletes images that are NOT used, so some were deleted. When I tried to export the ``imported`` entry, it could not export due to the same error.

Palloxin:
Export error:


It comes from an imported EPUB entry.
[922k.txt](https://github.com/user-attachments/files/17157552/922k.txt)
__note__: RENAME the extension from ``.txt`` to ``.epub`` (github doesnt accept epub files)
Import the epub, then try to export it.

Palloxin:
Flaw
Directory error after 2 min of exporting:

The problem is that I used a path that was picked in the past, and somehow it disappeared, so it gave the error.
Solution:
-Check the directory avaiability first, to make the error pop up immediately and not waste 2 minutes.

Palloxin:
Feature request:
Perma-edit chapters content with customJS upon export.
A way to edit the content of downloaded chapter based on customJS. A little checkbox somewhere here:
ii

Palloxin:
Suggestion:
Way to inject directly JS or CSS, ignoring the code in customJS and CSS of the app.
It's a little cumbersome to go there, paste the code and then come back to the entry to export.

It's more comfortable.

CD-Z:
> Suggestion: Way to inject directly JS or CSS, ignoring the code in customJS and CSS of the app. It's a little cumbersome to go there, paste the code and then come back to the entry to export.
> 
> It's more comfortable.

Why would you want that? What exactly is the use case to have js/css specific only for the epub. Especially when most epub readers don't support it.

Palloxin:
>Why would you want that? What exactly is the use case to have js/css specific only for the epub. Especially when most epub readers don't support it.

Just more comfortable. Since js can already be included, allow to directly do it from this place:
Screenshot_20241011-105549_1

Also, IF you have the intention of implementing the Perma-edit feature above, the "especially when most epub readers don't support it" wouldnt matter anymore.

CD-Z:
Well it won't come with this PR, for now I need to find the time to finish this one.

A perma edit would be the best solution to it, however this would also need a lot of work.

Palloxin:
Bug or Flaw: If in the destination folder there a file with the same name (and extension), the file will be overwritten (forced) by the latest one.

Add a (1) or something like that if that occours.

Palloxin:
Flaw: Export fails, ``Permission denied`` because storage permissions have not been given.

1. Upon export start make app.check if storage permissions have been given, IF NOT make a permission request window popup.

Palloxin:
Bug:
tried to export with beta export version:


-tried with ur latest epub export commit:

Of course, same entry

Palloxin:
Bug (or Flaw):

-Some special characters in the sumnary should probably be converted to html entities

result:


Used: ``@#€%-+&()/*"':;!?№$¢₱£¥‰℅—–_·±<[{]>}★„“”«»‚‘’‹›⋮¡¿‽…~\`|•√π÷×¶∆\=°^✓™®©⟩⟨♪ωαβΩΠμ§≠∞≈``

Resulted epub (rename the extension from ``.txt`` to ``.epub`` cuz github doesnt allow .epub ti be uploaded here): 
[Hidden_Marriage.txt](https://github.com/user-attachments/files/21212000/Hidden_Marriage.txt)

Ps: that same epub when imported on MoonReader simply does not display the epub summary, making it empty

EDIT: after some testing, 3 characters needed to be converted to be accepted by LNReader import and MoonReader Import (summary shows for moonreader now):
•these characters: ``&``, ``<``, and ``>``

Palloxin:
Bug
``<`` or ``>`` in chapter titles make the export produce a flawed epub (fake html tags).



After the export u can find the flaw like this (toc file):


It's a problem because even MoonReader simply deletes that chapter from the list when imported:



epub:
[FFFTier_System_SSSRank_Wife.txt](https://github.com/user-attachments/files/21976613/FFFTier_System_SSSRank_Wife.txt)
(*rename the file name extension to ``.epub`` because github doesn't allow ``.epub`` files to be uploaded*)


PS: that novel has been taken from NᴏvelFire source (en)
PSS: on LNReader, when imported on latest build, it simply renames all the chapter list with chapter0, chapter1, ignoring the toc

Palloxin:
Bug

If a chapter has an image, AND also a missing image(so 2 images), upon exporting, the app will fill that missing image with the one that exists, duplicating it.

On website (it's the same on LNReader):
https://github.com/user-attachments/assets/eca3e0aa-b1bd-4509-8bef-bb1c625202db

Imported epub (MoonReader):

https://github.com/user-attachments/assets/268f6cde-bb33-4661-a901-1deae054ccd3


Image that exists (1st):



Image that **doesnt** exist(2nd):



The plugin in question is ``CentralNovel``(portugese), ``Kono Subarashii``(novel), chapter is the 6th (prologue is chapter 1)

PS: I suggest to test on that source now, lest it disappear in the future and become untestable (idk where to find another source with same problem)

Palloxin:
Bug:

Export fails in this case: If a novel has a chapter, and that chapter has a broken image. Obviously the image isnt downloaded cuz it doesnt exist, but the html reference remains with the downloaded chapter. Then u try to export.
530804958-e143b9af-2e5c-4fe2-bfac-b4ae81e1f13e_1

Palloxin:
Feature request: Add export option to forcibly include the chapter number before the chap titles in the toc.

For the case where a novel in question has Zero chapter-number reference in the chap titles like here:

ghost:
Bter export when?


https://github.com/user-attachments/assets/8f76457d-3d51-4200-a2e1-1c829ebc0335

PR #1197, opened 2024-08-11
Progress bar - better proportions

Palloxin:
Since the tts buttons has been replaced by the bubble version, it's now possible.
Proportions taken based on Stable 1.1.19 except vertical (shorter than stable, longer than current beta)
If you want to make vertical like stable the height values is around 55vh
Current:
image
image

Modified (commit):
image
image

### DIFF
--- a/android/app/src/main/assets/css/index.css
+++ b/android/app/src/main/assets/css/index.css
@@ -160,7 +160,7 @@ td {
 #ToolWrapper > #ScrollBar {
   margin-top: 8px;
   width: 2.4rem;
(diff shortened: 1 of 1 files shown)


nyagami:
Im not OCD, but isnt 95.5 a not pretty number, the left and right space is 4.5/2=2.25 (2 number after decimal point)

Palloxin:
95 is fine too. I just reproduced the proportions from stable.

rajarsheechatterjee (approved):

PR #1164, opened 2024-07-27
Improve "Remove extra paragraph spacing" feature

Palloxin:
repost of #1105, closed by mistake.
Currently, it just removes all ``<br>``.
With this change you could:
-remove excessive br
-remove some zero-width spaces (I found a novel that put them in a p to simulate an extra empty paragraph).
-remove a br after the end of a p. It's just extra space.
-remove a br present right before the end of a p.

PR #1203, opened 2024-08-11
 Improve "Remove extra paragraph spacing" feature

Palloxin:
repost of #1164 due to conflicts (related code moved to another file)
Currently, it just removes all ``<br>``.
With this change you could:
-remove excessive ``<br>``
-remove some zero-width spaces and ``&nbsp;`` (I found a novel that put them in a p to simulate an extra empty paragraph).
-remove a br after the end of a p. It's just extra space.
-remove a br present right before the end of a p.

EDIT: the feature code has regex chaining now.

### DIFF
--- a/android/app/src/main/assets/js/core.js
+++ b/android/app/src/main/assets/js/core.js
@@ -465,7 +465,10 @@ window.addEventListener('DOMContentLoaded', async () => {
     }
 
     if (reader.generalSettings.val.removeExtraParagraphSpacing) {
-      html = html.replace(/<\s*br[^>]*>/gi, '\n').replace(/\n{2,}/g, '\n\n');
+      html = html
+        .replace(/([\u200B-\u200D\uFEFF]|&nbsp;)(?<=<p>\1)/g, '')
+        .replace(/(?:<br>\s*<\/?br>)+/g, '')
+        .replace(/<br>\s*(?=<\/?p>)/g, '');
     }
     reader.chapterElement.innerHTML = html;
   });
--- a/android/app/src/main/assets/js/core.js
+++ b/android/app/src/main/assets/js/core.js
@@ -467,7 +467,7 @@ window.addEventListener('DOMContentLoaded', async () => {
     if (reader.generalSettings.val.removeExtraParagraphSpacing) {
       html = html
         .replace(/(&nbsp;)(?<=<p[^>]*>\s*\1)/g, '')
-        .replace(/(?:<br>\s*<\/?br>\s*)+(?=<\/?(?:p|h[1-5])\b[^>]*>)/g, '')
+        .replace(/(?:<br>\s*<\/?br>\s*)+(?=<\/?(?:p|h[1-5]|br)\b[^>]*>)/g, '')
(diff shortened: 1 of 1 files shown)


nyagami:
<br> <p class="haha">
failed this test

android/app/src/main/assets/js/core.js
```diff
@@ -465,7 +465,10 @@ window.addEventListener('DOMContentLoaded', async () => {
     }
 
     if (reader.generalSettings.val.removeExtraParagraphSpacing) {
-      html = html.replace(/<\s*br[^>]*>/gi, '\n').replace(/\n{2,}/g, '\n\n');
+      html = html
+        .replace(/([\u200B-\u200D\uFEFF]|&nbsp;)(?<=<p>\1)/g, '')
```
nyagami, line 469: what is these kind of uncode. Do they make sense

android/app/src/main/assets/js/core.js
```diff
@@ -465,7 +465,10 @@ window.addEventListener('DOMContentLoaded', async () => {
...
+        .replace(/(?:<br>\s*<\/?br>)+/g, '')
```
nyagami, line 470: why adding `+` in the end. Isnt `g` flag is enough

android/app/src/main/assets/js/core.js
nyagami, line 469: are you sure `(?<=` would working properly
image

Palloxin: They do

Palloxin: With customJS the lookbehind works fine. I am sure if the regex of that feature has different flavour

android/app/src/main/assets/js/core.js
Palloxin: it's faster

android/app/src/main/assets/js/core.js
nyagami: Could you provide some examples.

android/app/src/main/assets/js/core.js
nyagami: You should be the one who provide a benchmark :).
But yeah, it maybe true

Palloxin: https://github.com/user-attachments/assets/84dea670-a2b7-4e11-86e6-4da85ad99d16

android/app/src/main/assets/js/core.js
nyagami: Why you have to use positive lookbeind and numeric reference instead of simple `<p>&nbsp;</p>`

Palloxin: because ``<`` are more frequent than & and zero width space. The regex will immediately fail without losing time.

Palloxin:
Double br exception:
cF6sm6l4
from
``/(?:<br>\s*<\/?br>\s*)+/``
to
``/(?:<br>\s*<\/?br>\s*)+(?![^<])/``

Palloxin:
I think it's ready now

Issue #1199, opened 2024-08-11
30° swipe angle thresold for browse section (plugin cards)


Palloxin:
### Describe your suggested feature

https://github.com/user-attachments/assets/ffac7e5b-8c43-4ca3-a1f3-2507b4e2e950

I can't scroll decently without being interrupted by the side options.

I suggest to implement a 30° swipe angle thresold on these plugin cards-
![immagine](https://github.com/user-attachments/assets/5019de54-cfd3-4133-bf58-1eddbe961ee3)

Scrolling the plugin list can be a bit annoying. A plugin card could be triggered horizontally as you can see in the video.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with a source, I should be opening an issue in the [sources repository](https://github.com/LNReader/lnreader-sources/issues/new/choose).
- [X] I have updated the app to version **[2.0.0](https://github.com/LNReader/lnreader/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

Palloxin:
No need for this anymore

PR #1219, opened 2024-08-17
Enhancement: Make swiping of plugins/sources less sensitive

Skeltons:
Make swiping source/plugins less sensitive, related issue #1199 

https://github.com/user-attachments/assets/de4136f1-1b25-4f5f-b869-43b1befae593

### DIFF
--- a/src/screens/browse/components/BrowseTabs.tsx
+++ b/src/screens/browse/components/BrowseTabs.tsx
@@ -80,6 +80,8 @@ export const InstalledTab = memo(
       ({ item }) => {
         return (
           <Swipeable
+            dragOffsetFromLeftEdge={62}
+            dragOffsetFromRightEdge={62}
             renderLeftActions={(progress, dragX, ref) => {
               return (
                 <View


rajarsheechatterjee (approved):

Palloxin:
https://github.com/LNReader/lnreader/issues/1199
this would help too

Palloxin:
Btw 62 is too much. It's cranky now. Can u try 40?

Skeltons:
I have tried 50 and it was still triggering the swipe, i think this would be quite hard to balance

Palloxin:
In ny opinion it's enough to lower the sensivity to 30-40 reducing the issue until a 30° angle thresold is implemented

Skeltons:
It is currently hard coded but the best way is probably to turn it into a function of your phone's width

Skeltons:
If you dont mind can i have your phone model

rajarsheechatterjee:
In my opinion, we shouldn't adjust its value based on lower-end or older devices. If it's working well for most devices, that's sufficient. He uses a Huawei P8 Lite 2017.

Palloxin:
My device is very smooth when using plugin cards, it's not due to that.

Skeltons:
> My device is very smooth when using plugin cards, it's not due to that because there is no lag, surprisingly.

Ah its a misunderstanding i wanted to know your device so i can get the width and find a correlation

Palloxin:
1080x1920. Maybe a % value would be better?

CheshireCat27:
Tested on a Moto One 5G Ace, was able to get minor hitching when installing plugins, but nothing else.

CheshireCat27:
I have a painfully slow tablet I can test on as well if it would be helpful.  However it barely runs a handful of Firefox tabs so I wouldn't say its indicative of the average low end device.

Palloxin:
Here is my data, I used an autoclicked with precise coordinates, Macrorify.
-720x1280 (I changed it through settings)
-speed: 20

Value 35 (minimum non-responsive width: 65px)

https://github.com/user-attachments/assets/847b1b55-92ea-469c-8926-c7bbef435916

Value 62 (minimum non-responsive width: 115px)

https://github.com/user-attachments/assets/224b558c-4a34-437b-b735-2eba36a15efe

Palloxin:
Practical test videos:
Value: 30
Comfortable, no unwanted trigger

https://github.com/user-attachments/assets/3adea854-c765-4ee5-8f6b-604e68a244e0

Value 25
Some unwanted trigger start to appear. 30 os better.

https://github.com/user-attachments/assets/424b6a04-596e-4183-b7a5-4fd295bc6fb1

Palloxin:
The default value was 10(before this commit). Three times it reduces the problems by 90%.
Stopping completely unwanted triggers comes with the price of cranky plugin cards.
It's better to just reduce them.

Skeltons:
Ok that sounds like a good compromise, pretty unrealistic to expect no triggers

PR #1220, opened 2024-08-17
Update swiping offset to be less aggressive

Skeltons:
Feedback from other users seems to be that the swiping offset is too aggressive, making swiping left and right too janky.
Lowered the value from 62 to 30, triggering the swipe would be happen more frequently when scrolling, but swiping left and right would be much smoother.


Palloxin:
Can we have 28? 🥺

Skeltons:
I know I wrote 62 in the previous pull request but ideally, I wanted a nice number in the tens 😄

rajarsheechatterjee (approved):

PR #1225, opened 2024-08-20
Re-disable user-select on next-button and info-text

Palloxin:
After VanJS commit it was disabled by mistake.


nyagami:
Could you add it for next button too?

Palloxin:
@nyagami 
Screenshot_20240821-114746_1
alredy done

PR #1266, opened 2024-10-02
fix 'cant parse page' sql syntax error

nguyd1:
had issue on novelupdates where opening novels for the first time will give cant parse page error since sql has wrong syntax


Palloxin:
He talking about this 
EZJjOpjx

rajarsheechatterjee (approved):

PR #1296, opened 2024-10-30
Fix problem with android nav bar on A8 miui 8

Palloxin:
There is a harassing bug on Android 8 miui 8 with android navigation bar. 

This:

https://github.com/user-attachments/assets/a4b0da33-641e-4205-a048-6062d1de346d

The pr solution comes from nyagami, and works. 

I dont have the ability to test on other devices tho. Feel free to test if there are problems. So far in 2-3 i tested i never encountered a problem.

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -1,5 +1,6 @@
 import 'react-native-gesture-handler';
 import { enableFreeze } from 'react-native-screens';
+import * as NavigationBar from 'expo-navigation-bar';
 
 enableFreeze(true);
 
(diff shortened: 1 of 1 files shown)


Palloxin:
the ONLY flaw I found is that the bug re-appears if a certain action is reproduced:

https://github.com/user-attachments/assets/5a1c309f-dd43-47db-8664-bdd0db31a8a6

1. Open app, open a random chapter
2. android nav bar works fine, bug fixed
3. Click back button until app closes itself 
4. Click app icon again
5. Open a random chapter

The bug is there again until I close the app in the normal way.

PR #1314, opened 2024-11-06
Add 'multi' language option

NilSilva:
Add option for plugins with no language. They don't have a specific language so it shouldn't go to a specific language. I'm doing this with a komga plugin in mind.


src/utils/constants/languages.ts
```diff
@@ -18,6 +18,7 @@ export const languagesMapping: Record<string, string> = {
   'zh': '中文, 汉语, 漢語',
   'ja': '日本語',
   'ko': '조선말, 한국어',
+  'multi': 'Multi',
```
nyagami, line 21: should be a translated text

rajarsheechatterjee (approved):

nyagami (approved):

PR #1320, opened 2024-11-14
Small changes following faster global search pr

Soopyboo32:
What the colour looks like for loading stuff:
image


src/hooks/persisted/useSettings.ts
```diff
@@ -163,7 +163,7 @@ const initialAppSettings: AppSettings = {
 const initialBrowseSettings: BrowseSettings = {
   showMyAnimeList: true,
   showAniList: true,
-  globalSearchConcurrency: 1,
+  globalSearchConcurrency: 3,
```
Soopyboo32, line 166: most phones should be able to handle 3 and the first few concurrency increases makes a massive difference in loading time so may as well increase default

src/utils/useLoadingColors.ts
```diff
@@ -0,0 +1,34 @@
+import { ThemeColors } from '@theme/types';
+import color from 'color';
+import { useAppSettings } from '@hooks/persisted';
+import { interpolateColor } from 'react-native-reanimated';
+
+const useLoadingColors = (theme: ThemeColors) => {
```
Soopyboo32, line 6: renamed to use cus of the useAppSettings inside

src/screens/novel/components/LoadingAnimation/NovelScreenLoading.tsx
```diff
@@ -68,7 +68,7 @@ const NovelInformation = memo(() => (
   </View>
 ));
 
-const ChapterItem = memo(({ index }: { index: number }) => (
+const ChapterItem = memo(() => (
```
Soopyboo32, line 71: lint was mad at unused variable so i removed it

src/utils/useLoadingColors.ts
Palloxin: maybe lint gives problems because Traitor latest commit bypassed lint verification.

src/screens/novel/components/LoadingAnimation/NovelScreenLoading.tsx
Palloxin: maybe lint gives problems because Traitor latest commit bypassed lint verification.

nyagami (approved):

Palloxin:
Screenshot_4_1

I dont think it's good looking, why blue?

Soopyboo32:
> I dont think it's good looking, why blue?

its just a blend of the existing colors 
image
maybe those are blueish

PR #1365, opened 2024-12-27
CREATE BACKUP FILE NOT WORKING, FIXED

MOA-YAD:
in android file system you can't use the char  ' : '  so lnreader file name:
"Inreader_backup_YYYY-MM-DD_HH:mm.json"
you can't write it as a file. 
so i edit it to looks like
"Inreader_backup_YYYY-MM-DD_HH_mm.json"

### DIFF
--- a/src/services/backup/legacy/index.ts
+++ b/src/services/backup/legacy/index.ts
@@ -21,7 +21,7 @@ export const createBackup = async () => {
     if (!folder) {
       return;
     }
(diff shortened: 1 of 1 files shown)


MOA-YAD:
what i know is that you cant create a file contains : in his name 
my device is android 14 and i can't use :

i try it in my another device is android 8 and it worked 
so in new version of android you can't use :

Screenshot_20241227-140308_Settings

nyagami (approved):

Issue #676, opened 2023-06-29
Bookmark icon goes back to unbookmarked state(aesthetically)


Palloxin:
### Steps to reproduce

https://github.com/LNReader/lnreader/assets/75091899/0dd28d9b-49e2-4245-a1a5-754602982866

### Expected behavior

Bookmark icon doesnt reset automatically.

### Actual behavior

bookmark icon resets

### LNReader version

1.1.15.1

### Android version

android 8 miui 8

### Device

Huawei p8 lite 2017

### Other details

The chapter bookmarked in this way is in fact really bookmarked. It is just the icon that doesn't reflect the bookmarking state of the chapter.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with a source, I should be opening an issue in the [sources repository](https://github.com/lnreader/lnreader-sources/issues/new/choose).
- [X] I have updated the app to version **[1.1.15](https://github.com/lnreader/lnreader/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

PR #1368, opened 2024-12-28
Fixed last read not updating and bookmark button in reader mode.

jacksin125:
Previously, the bookmark button in the reader top bar would always reset to whatever the status was when you first opened the chapter. Additionally, if you read part of a chapter or bookmarked a chapter, left reader mode, then pressed the "continue reading" button to get back to where you left off, the bookmark status and reading progress would not have been saved like if you opened the chapter via the list.

I tried to make the least intrusive changes I could, since the core problem with the "continue reading" bug is that state variables are not updated, and instead changes are directly saved to the database in other sections. The best solution would require rewriting a good section of the codebase.

If changes are necessary, please let me know.

### DIFF
--- a/src/screens/reader/ReaderScreen.tsx
+++ b/src/screens/reader/ReaderScreen.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useCallback } from 'react';
+import React, { useRef, useCallback, useState } from 'react';
 import { DrawerLayoutAndroid } from 'react-native';
 
 import { useChapterGeneralSettings, useTheme } from '@hooks/persisted';
@@ -63,6 +63,7 @@ export const ChapterContent = ({
   const readerSheetRef = useRef<BottomSheetModalMethods>(null);
   const theme = useTheme();
   const { pageReader = false, keepScreenOn } = useChapterGeneralSettings();
+  const [bookmarked, setBookmarked] = useState(chapter.bookmark);
 
   const {
     hidden,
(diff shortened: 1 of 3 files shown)


nyagami:
Hi, could you provide steps or related issues for `the bookmark button in the reader top bar would always reset to whatever the status`

jacksin125:
1. Open a chapter in the reader. You _might_ need to use the continue reading button, but I don't think you do.
2. Tap the screen so the top and bottom bar appear.
3. Tap the bookmark button in the top right to change the bookmark status.
4. Tap the screen so the top and bottom bar disappear.
5. Make the top and bottom bar appear again, the bookmark icon will have reset.

This issue is caused due to the bookmark status being reset based on the chapter state variable every time the bar is reopened. However, the chapter state variable is used throughout so much of this section of the app that changing it forces a complete reload of the reader. So, I just moved the code up to the reader itself and passed the relevant variables to the reader bar, so they are not affected by the reader bar reopening.

Palloxin:
I once reported the same bug, and was fixed. It came back it appears https://github.com/LNReader/lnreader/issues/676

nyagami:
try using `setChapter` from `useChapterContext`  in `ReaderAppbar` to update chapter state instead. bookmarked state is no longer neccessary

jacksin125:
> try using `setChapter` from `useChapterContext` in `ReaderAppbar` to update chapter state instead. bookmarked state is no longer neccessary

I tried this, but due to where and how the chapter state is used, updating it forces a full reload of the reader and a refetch of the chapter. All the truly good solutions would require changing quite a bit more of the codebase to work properly, this was the best solution I came up with to avoid having to completely change how things are done in other sections.

nyagami (approved):

PR #1390, opened 2025-01-04
Arabic rtl

Soopyboo32:


src/screens/reader/components/WebViewReader.tsx
```diff
@@ -175,6 +177,16 @@ const WebViewReader: React.FC<WebViewReaderProps> = ({
...
+              ${
+                pluginInfo?.lang === '‎العربية'
+                  ? `
+                body {
+                  direction: rtl;
+                  text-align: right; 
+                }
+              `
+                  : ''
+              }
```
Soopyboo32, line 189: :waytoodank: not sure if i like how prettier formats this situation

nyagami: is only العربية rtl?

Soopyboo32: no clue cc @K1ngfish3r

K1ngfish3r: yes, none of the other languages currently supported use rtl

nyagami: Oh, we forgot multi lang (komga)

K1ngfish3r: For komga, that might require checking the html for a `dir` class
that would definitely be more cleaner than the current solution, but also involves the website to actively advertise it's use

nyagami: https://stackoverflow.com/questions/26549354/android-determine-if-device-is-in-right-to-left-language-layout

Havent read the article carefully, but it might help

K1ngfish3r: the first answer is promising(?). add a file that gets read from plugin side that would let the app side determine if RTL needs to be applied. I'm not sure how the multisrc side would work though.

the other solutions below it mainly determine the app's locale and apply RTL. Most likely will convert LTR plugins to RTL as well if device is in an arabic/rtl language

rajarsheechatterjee:
Closing this PR as it appears outdated.

If the changes are still relevant, feel free to reopen the PR or submit a new one.

*Note: This was done as part of an automated cleanup of stale pull requests.*

PR #1395, opened 2025-01-18
Fix app crash when downloading novel entry if it is on the recent updates page

Soopyboo32:


Soopyboo32:
not that i know of, will double check now gimme like 3mins

Soopyboo32:
ye can confirm no more errors

nyagami (approved):

PR #1412, opened 2025-01-27
Ukrainian language

db4tz:
Translated all lines from uk_UA/strings.json to Ukrainian. Tested it on my build, works fine. If I missed something let me know.


CD-Z:
Normally translation should be added through [crowdin](https://crowdin.com/project/lnreader). Don't know if it works this way or creates problems. @rajarsheechatterjee

db4tz:
> Normally translation should be added through [crowdin](https://crowdin.com/project/lnreader). Don't know if it works this way or creates problems. @rajarsheechatterjee


Oh, I didn't know. I can copy everything to crowdin. It's just that with json editing I could test it right away.

PR #1414, opened 2025-01-31
Update global search result if added to library

Soopyboo32:
I added usememo in better globalsearch pr

### DIFF
--- a/src/screens/GlobalSearchScreen/components/GlobalSearchResultsList.tsx
+++ b/src/screens/GlobalSearchScreen/components/GlobalSearchResultsList.tsx
@@ -157,7 +157,7 @@ const GlobalSearchSourceResults: React.FC<{ item: GlobalSearchResult }> = ({
         </View>
       </>
     ),
-    [item.isLoading],
+    [item.isLoading, library],
   );
 };
 


src/screens/GlobalSearchScreen/components/GlobalSearchResultsList.tsx
```diff
@@ -157,7 +157,7 @@ const GlobalSearchSourceResults: React.FC<{ item: GlobalSearchResult }> = ({
         </View>
       </>
     ),
-    [item.isLoading],
+    [item.isLoading, library],
```
nyagami, line 160: I dont see `library` was used in useMemo body

Soopyboo32: image
here ^
image

Soopyboo32: was hidden xd

nyagami: Is using `library` as the dependency instead of `novelInLibrary` better?

Soopyboo32: well novelinlibrary changes every render
so would need to wrap novelinlibrary with a useCallback dependant on library if we did that
and it would be the same thing in the end xd
(idk maybe that would be better best practice for react tho?)

PR #1458, opened 2025-03-16
Faster import + some regex improvements

Palloxin:
Faster import by improving its regex.
In addition, improved others too without changing logic


src/utils/parseChapterNumber.ts
```diff
@@ -9,11 +9,11 @@ export const parseChapterNumber = (
     return chapterNumber;
   }
 
-  const basic = new RegExp(/(?<=ch[^\d]*[\s]*)([0-9]+)(\.[0-9]+)?(\.?[a-z]+)?/);
-  const number = new RegExp(/([0-9]+)(\.[0-9]+)?(\.?[a-z]+)?/);
+  const basic = new RegExp(/(\d+)(?<=ch\D*\s*\d+)(\.\d+)?(\.?[a-z]+)?/);
```
Palloxin, line 12: never start a regex with lookbehind lookahead.

Issue #1470, opened 2025-04-07
Back the top in chapter list


CoffeePython95:
### Describe your suggested feature

Please add button, that will move users view to the top in chapters list.
It would be nice if it was small pop-up button, that appears only when user scrolls a bit into the list.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] If this is an issue with a source, I should be opening an issue in the [sources repository](https://github.com/LNReader/lnreader-sources/issues/new/choose).
- [x] I have updated the app to version **[2.0.0](https://github.com/LNReader/lnreader/releases/latest)**.
- [x] I will fill out all of the requested information in this form.

PR #1479, opened 2025-04-18
feat: Back the top in chapter list (#1470)

Neel-shetty:
Addresses #1470
Added a scroll to top button in the novel screen
Aligned it to the left because resume FAB takes the button right space when enabled
Let me know if I need to make any changes
<div align="center">   </div>


src/screens/novel/NovelScreen.tsx
```diff
@@ -499,6 +505,21 @@ const Novel = ({ route, navigation }: NovelScreenProps) => {
...
+          {showScrollToTop ? (
+            <AnimatedFAB
+              style={[
+                styles.scrollTopFab,
+                { backgroundColor: theme.primary, marginBottom: bottomInset },
+              ]}
+              extended={isFabExtended}
+              color={theme.onPrimary}
+              uppercase={false}
+              label={getString('readerScreen.drawer.scrollToTop')}
```
nyagami, line 517: I dont think we need label for this action

Neel-shetty:
Done, kept the empty string because label is a required prop.

CD-Z:
Personally I think it would be better to have another way to display this. Scroll to top is not a primary action, so it shouldn't be the primary color. 
Also wouldn't it be better to have the buttons on one side? It feels weird to have buttons on the bottom left.

So my suggestion would be to put the button above the other FAB (if it is activated) and color it in a secondary color.

Palloxin:
We just need a scroll bar :)

Neel-shetty:
> Personally I think it would be better to have another way to display this. Scroll to top is not a primary action, so it shouldn't be the primary color. Also wouldn't it be better to have the buttons on one side? It feels weird to have buttons on the bottom left.
> 
> So my suggestion would be to put the button above the other FAB (if it is activated) and color it in a secondary color.

I tried adding it above the resume button but it looked a bit awkward there.
Tried hiding the resume when you scroll a bit down and only show scroll to top button on the right, it works fine but some users might want to resume after scrolling a few chapters I'm not sure. 
Scroll bar sounds like a good idea.
I went with the button as it solved my issue and I saw an open issue with the feature request tag.

@nyagami would like your opinion on what I should implement.

Issue #1454, opened 2025-03-02
[Flaw] Re-add the ability to save entry covers


Palloxin:
### Describe your suggested feature

In beta the ability to save covers is absent. Re-add this feature.

![Image](https://github.com/user-attachments/assets/cb0abbae-34d2-4a94-b384-9fbad172730b)

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] If this is an issue with a source, I should be opening an issue in the [sources repository](https://github.com/LNReader/lnreader-sources/issues/new/choose).
- [x] I have updated the app to version **[2.0.0](https://github.com/LNReader/lnreader/releases/latest)**.
- [x] I will fill out all of the requested information in this form.

PR #1481, opened 2025-04-22
feat: Save novel cover (#1454)

Neel-shetty:
Addresses #1454 
Added save novel cover button

Let me know if i should add it in the NovelAppBar as well since Edit cover option is also present over there


Palloxin:
What happens if user didnt give storage permissions?
If missing it has ti ask user to give storage permissions

Neel-shetty:
okay thank you, i'll look into it

Neel-shetty:
used Storage Access Framework to get the permission for saving the image, works fine on release build now.
previous method worked fine in dev but crashed in release as you mentioned.

Neel-shetty:
updated the code for new arch

rajarsheechatterjee (approved):

PR #1488, opened 2025-05-04
fix: fix alignment of save novel cover button

Neel-shetty:
Save and edit buttons for novel cover image were misaligned while resolving conflicts

Before:

After:


Palloxin:
@Neel-shetty 
Screenshot_20250509-222420_1
The two buttons arent.. themed. u cant distinguish them from the ones behind.

Palloxin:
@Neel-shetty 
Screenshot_20250511-102618_1
Hmm, the app doesnt 'ask' for storage permissions if the latter is missing. It just gives a message error.

Issue #1276, opened 2024-10-12
Legacy Backup is not working 


Pheonix-Flames:
### Steps to reproduce

Go to More > Settings > Backup > Create Backup 
Try to create a backup 

### Expected behavior

It should create a backup in the selected location 

### Actual behavior

It is not creating a backup and there is an error poping up
![Screenshot_2024-10-12-20-55-06-63_3c1c179116822eec69c21805387f0489](https://github.com/user-attachments/assets/4bd4744f-db81-4081-a930-dc74b719f3b8)

### LNReader version

2.0.0

### Android version

Android 11

### Device

Realms 3 Pro 

### Other details

Please solve this issue I have to transfer my Data to other phone

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with a source, I should be opening an issue in the [sources repository](https://github.com/lnreader/lnreader-sources/issues/new/choose).
- [X] I have updated the app to version **[2.0.0](https://github.com/lnreader/lnreader/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

nyagami:
Delete the folder and try again

Pheonix-Flames:
> Delete the folder and try again

Any Updates on this problem ?

ilonachan:
I've implemented a solution by replacing the current hack of "select a folder for full write permissions and then copy into it" (which evidently doesn't work anymore) with a proper file export intent using `@react-native-documents/picker`.

**Compatibility issue:** The library is cross-platform, but the URI handling is probably different on iOS and I can't test it.

I also essentially rewrote the entire legacy logic to replicate what the "Self-Hosted" backup does, except instead of uploading two zip files, I just combine them into one and export it (for now just as a zip file, but calling it something like `.lnrbak` could be a good idea).

Note that I've simply copied the logic of self-hosted backups exactly; I have not fixed any existing bugs in that logic, some of which I've noticed and will look into further. Both (or all three) backup types use the same utility methods, so their functionality and correctness is now DRYly coupled.

Since this is such a fundamentally basic feature, I believe writing it off as "legacy" is really bad, and it'd be good if that wording could be removed in the translation strings. "Local Backup" is more descriptive.

ovizii:
[...]

PR #1603, opened 2025-10-15
Reimplemented Local Backup based on Self-Hosted backup  #1276

ilonachan:
Exactly what it says on the tin. The logic for gathering data in a "data.zip" and "download.zip" file is the same as for self-hosted (and GDrive?) Backup, the only difference is that I place "download.zip" inside the other archive to export both as a single zip file. Import works too, to the same degree that it works for the other methods.


ilonachan:
If this goes through, I would also like the translation strings referring to "Legacy Backup" to be changed to "Local Backup", since that better describes the purpose and importance of this feature.

Palloxin:
Tested. It does work. It creates the backup, but after wiping the data and trying to restore the backup on the very same build, tgere is no hint of backup in progress or completed. Add the restoring process to the download queue. To see the backup restored u need to restart the app.

Also, either force an app restart after backup restoring is completed, or find a way to make the restored changes apply immediately everywhere. The first solution seems much easier.

Come to the discord server if u want to discuss.

ilonachan:
I see, these are all great points. So I should probably dig deeper into how the other backup mechanisms handle background tasks and app reloading.

> Come to the discord server if u want to discuss.

I do want to, but I won't have time consistently. No harm in joining tho

PR #1620, opened 2025-10-22
Plugin repo API version check & warning

ilonachan:
This is a future-proofing measure: if, according to my proposal, an `api` version field is attached to plugins to manage breaking changes to the plugin API, older versions of the app should be able to detect and reject plugins for newer API versions. No sophisticated logic is required for now: as long as the very first stable 2.0 release has this feature, users won't be confused once plugin version starts actually being used later.


rajarsheechatterjee:
Closing this PR as it appears outdated.

If the changes are still relevant, feel free to reopen the PR or submit a new one.

*Note: This was done as part of an automated cleanup of stale pull requests.*

PR #1668, opened 2025-11-27
Update Chapter loading behaviour

CD-Z:
Fixes the issues regarding the chapter loading behavior in JumpToChapter and the Chapter Drawer.

Branch based on #1667


Palloxin:
what about moving the "go to top" button inside "Jump To" section since you are already there? Frankly that button is a punch on the face as it is.
FjXlVxK6

CD-Z:
No, this isn't what this PR is about.

PR #1685, opened 2025-12-16
Add volume buttons offset setting and fix rlated ui garbage

### LINKED ISSUE #1507: Volume buttons scroll offset value not showing
### Steps to reproduce

Enable volume buttons scroll

### Expected behavior

There should be a box below the option to change the offset value 

### Actual behavior

Nothing shows up

### LNReader version

2.0.0

### Android version

Android 14

### Device

Resmi 8A

### Other details

İn the old version (1.9) I could change the offset value of the volume buttons scroll but even on the old version it was still a bit buggy the box would only appear if I open a novel then go to settings from the menu below and then turn my phone to landscape mode. İn the 2.0.0 version even if I turn my phone to landscape mode it doesn't show up. Only once it appeared on the normal settings menu but after reinstalling the app it doesn't appear at all (autoscroll scroll offset does show up if I enable it but it doesn't effect the volume buttons scroll offset value)

### Acknowledgements

- [x] If this is an issue with a source, I should be opening an issue in the [sources repository](https://github.com/lnreader/lnreader-sources/issues/new/choose).
- [x] I have updated the app to version **[2.0.0](https://github.com/lnreader/lnreader/releases/latest)**.

686udjie:
closes #1507

### DIFF
--- a/src/screens/settings/SettingsReaderScreen/tabs/NavigationTab.tsx
+++ b/src/screens/settings/SettingsReaderScreen/tabs/NavigationTab.tsx
@@ -12,6 +12,7 @@ const NavigationTab: React.FC = () => {
   const theme = useTheme();
   const {
     useVolumeButtons = false,
+    volumeButtonsOffset = null,
     verticalSeekbar = true,
     swipeGestures = false,
     pageReader = false,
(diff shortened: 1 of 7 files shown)


src/screens/settings/SettingsReaderScreen/tabs/NavigationTab.tsx
```diff
@@ -43,6 +44,28 @@ const NavigationTab: React.FC = () => {
           }
           theme={theme}
         />
+        {useVolumeButtons && (
+          <View style={styles.inputContainer}>
+            <TextInput
+              label={getString('readerSettings.volumeButtonsOffset')}
```
686udjie: for this, the naming is kinda scuffed. any ideas for what it should be called?

rajarsheechatterjee (approved):

PR #1747, opened 2026-02-15
Improves library update concurrency and TTS notification handling

Prime-AI28:
Speeds up library updates by batching chapter DB operations and running source updates in parallel, reducing lock errors and UI jank. Switches to a read-only DB connection for safe concurrent reads. Refactors TTS notification actions to use an event bus, enabling instant response to user controls. Adds smarter update delays and prioritizes ongoing novels for efficient background processing.

Enhances error handling and prevents crashes from failed file deletes or async actions. Refines notification buttons for accurate Play/Pause state. Updates library screen logic to reduce contention and improve perceived responsiveness.

Fixes rare database lock errors and makes novel updates more reliable and user-friendly.

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -78,8 +78,8 @@ const App = () => {
       response => {
         const actionIdentifier = response.actionIdentifier;
         if (actionIdentifier.startsWith('TTS_')) {
-          const { setTTSAction } = require('@utils/ttsNotification');
-          setTTSAction(actionIdentifier);
+          const { emitTTSAction } = require('@utils/ttsNotification');
+          emitTTSAction(actionIdentifier);
         }
       },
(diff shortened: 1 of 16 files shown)


CD-Z:
You added a mihon and opt_mod file, which both are useless and should not be committed. The change to open two db connections is a nice idea, but doesn't help in practice as both connections try to access a single sqlite file and will still block each other.

Prime-AI28:
Hey sorry for that mistake about mihon and opt_mod one, and thanks for pointing the shortcomings, i have started using this app and liked it quite a bit faced multiple crashes and just wanted to see if i could change/improve it a bit. I will keep theses points in mind for future.

PR #1764, opened 2026-02-27
Feat/tts

Yashwanth12321:


CD-Z:
We don't need the .planning directory of your agent. Also please provide an description what you did, what the goal of this PR is and how it improves the current TTS. It would be also great if you create unit tests.

PR #1797, opened 2026-03-25
fix: Apply active filters to chapter count display (Closes #1793)

### LINKED ISSUE #1793: Chapter Count Not Updated After Filtering Unread Chapters
### Steps to reproduce

1. Open any Novel from the Library
2. Apply filter to show only unread chapters
3. Observe the chapter count displayed

### Expected behavior

The total chapter count should update dynamically to reflect only the currently displayed chapters (i.e., unread chapters after filtering).

### Actual behavior

The chapter count remains unchanged and continues to show the total number of chapters (read + unread), even though only unread chapters are displayed.

### LNReader version

2.0.2

### Android version

Android 11

### Device

Realme 3 Pro

### Other details

This leads to inconsistent UI behavior, as the visible chapter list does not match the reported total, which may confuse users when tracking reading progress.

### Acknowledgements

- [x] If this is an issue with a plugin, I should be opening an issue in the [plugins repository](https://github.com/lnreader/lnreader-plugins/issues/new/choose).
- [x] I have updated the app to version **[2.0.3](https://github.com/lnreader/lnreader/releases/latest)**.

Definisi:
Fixes #1793

When filtering chapters (e.g. showing only unread), the chapter count in the novel header kept showing the total instead of the filtered count.

**Root cause:**
`getChapterCount` in `ChapterQueries.ts` only accepted `novelId` and `page`, with no way to pass the active filter. Meanwhile `getPageChaptersBatched` (which fetches the actual chapter list) was correctly applying filters via `chapterFilterToSQL`.

**What changed:**
- Added an optional `filter` parameter to `getChapterCount` and applied `chapterFilterToSQL` in the count query
- Updated both call sites in `useNovel.ts` to pass the current `settingsFilter`

The reader's `useChapter.ts` also calls `getChapterCount` (to check if a next/prev page has chapters), but those calls intentionally stay unfiltered since the reader should still allow navigating across all chapters.

**Testing:**
- All 31 existing tests pass
- No new type errors introduced

### DIFF
--- a/src/database/queries/ChapterQueries.ts
+++ b/src/database/queries/ChapterQueries.ts
@@ -395,10 +395,18 @@ export const getPageChapters = async (
   return query.all();
 };
 
-export const getChapterCount = async (novelId: number, page: string = '1') =>
+export const getChapterCount = async (
+  novelId: number,
+  page: string = '1',
+  filter?: ChapterFilterKey[],
+) =>
   await dbManager.$count(
     chapterSchema,
-    and(eq(chapterSchema.novelId, novelId), eq(chapterSchema.page, page)),
+    and(
+      eq(chapterSchema.novelId, novelId),
+      eq(chapterSchema.page, page),
+      chapterFilterToSQL(filter),
+    ),
   );
 
 export const getPageChaptersBatched = async (
--- a/src/database/queries/__tests__/ChapterQueries.test.ts
+++ b/src/database/queries/__tests__/ChapterQueries.test.ts
@@ -712,6 +712,54 @@ describe('ChapterQueries', () => {
 
       expect(result).toBe(2);
     });
+
+    it('should return filtered count when filter is provided', async () => {
+      const testDb = getTestDb();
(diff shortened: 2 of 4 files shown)


CD-Z:
Please fix the lint warnings in useNovel and tests for the filter behavior.

Definisi:
Fixed in e05a70b9:

- **Lint**: the `react-hooks/exhaustive-deps` warning on `getChapters` was because `novelSettings.filter` was used on line 274 instead of `settingsFilter` (same value, different reference). Swapped to `settingsFilter` so it matches the dep array.

- **Tests**: added two test cases for `getChapterCount` with filters:
  - `should return filtered count when filter is provided` (verifies `not-read` and `read` filters return correct counts)
  - `should return total count when no filter is provided` (verifies `undefined` and `[]` both return total)

  Also updated `chapterFilterToSQL` mock in `setup.ts` to actually apply SQL conditions instead of always returning `undefined`, so the filter tests can work.

All 201 tests passing.

rajarsheechatterjee (approved):

CD-Z (approved):

PR #1804, opened 2026-04-02
Expo 55 upgrade

CD-Z:


Palloxin:
sppt
 

legenlist => legendlist 

there are two of these typos in the code

Palloxin:
the build didnt buuld

CD-Z:
Yeah, they removed the Hermes compiler from the react native package into its own package. I forgot about it. I added it and am just building locally to see if it runs through now.

PR #1824, opened 2026-05-02
feat(plugins): add Cloudflare fetch fallback and cookies API

tinywind:
## Summary

Plugins frequently target Cloudflare-protected sites, and the JS fetch implementation on Android cannot survive interstitial challenges. This PR adds two complementary pieces:

1. **Cloudflare fallback in `fetchApi`** ([src/plugins/helpers/fetch.ts](https://github.com/lnreader/lnreader/blob/master/src/plugins/helpers/fetch.ts)) — when a `403` or `503` `text/html` response matches a Cloudflare challenge body (`cf_chl_opt`, "Just a moment...", `challenge-platform`, `cf-mitigated`), retry via `NativeFile.downloadFile` (`webFetch`). The platform HTTP stack tends to clear most challenges. On success the retry response is returned; otherwise the original body is preserved with the original status, so callers see no regression in the failure path.
2. **`@libs/cookies` API for plugins** ([src/plugins/pluginManager.ts](https://github.com/lnreader/lnreader/blob/master/src/plugins/pluginManager.ts)) — expose `get` / `set` / `clearByName` via `@preeternal/react-native-cookie-manager`, backed by the shared WebView/HTTP cookie store. Pairs with the Cloudflare retry so plugin code can capture session cookies and reuse them on subsequent same-origin requests.

## Why

A growing number of source sites sit behind Cloudflare. Currently those plugins fail with the challenge HTML body, with no recourse for users. The fallback uses the Android OkHttp-backed downloader which respects the same WebView cookie jar that already authenticates the user, so a single round trip is usually enough to clear the interstitial.

## Test plan
- [ ] Install a plugin pointing at a Cloudflare-protected source.
- [ ] Confirm fetching the homepage / a chapter previously failing with 403/503 now returns content.
- [ ] In a plugin body, `const { cookies } = require('@libs/cookies')` and verify `get/set/clearByName` round-trip a cookie via the WebView store.
- [ ] Non-Cloudflare requests take the fast path (no observable behavior change for normal 2xx responses).

## Notes
- Detection uses a fixed regex over the response body; if Cloudflare changes its boilerplate the regex may need updating.
- `webFetch` writes to `ExternalCachesDirectoryPath` (or `/data/local/tmp` fallback) and `unlink`s the temp file in `finally`.
- No new dependency added beyond `@preeternal/react-native-cookie-manager`, which is already declared in `package.json`.

### DIFF
--- a/src/plugins/helpers/fetch.ts
+++ b/src/plugins/helpers/fetch.ts
@@ -40,12 +40,98 @@ const makeInit = (init?: FetchInit) => {
   return init;
 };
 
+const headersToObject = (
+  headers?: Record<string, string> | Headers,
+): Record<string, string> => {
+  const out: Record<string, string> = {};
+  if (!headers) {
+    return out;
+  }
+  if (headers instanceof Headers) {
+    headers.forEach((value, key) => {
+      out[key] = value;
+    });
+    return out;
+  }
+  for (const [key, value] of Object.entries(headers)) {
+    if (typeof value === 'string') {
+      out[key] = value;
+    }
+  }
+  return out;
+};
+
+const isCloudflareChallengeBody = (body: string): boolean =>
+  /Just a moment\.\.\.|cf_chl_opt|challenge-platform|cf-mitigated/i.test(body);
+
+const cacheDir =
+  NativeFile.getConstants().ExternalCachesDirectoryPath ?? '/data/local/tmp';
+
+let webFetchSeq = 0;
+
+const webFetch = async (
+  url: string,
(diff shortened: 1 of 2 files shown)


tinywind:
Closing — wrong target. The intended fix is the gesture-handler Pressable change in commit 6f6b3132 (resolves the WSA Installed-plugin click bug). Reopening as a separate PR; the cookies / Cloudflare-fallback work in this branch will live in tinywind/lnreader for now and may be re-proposed later if there is interest.

PR #1831, opened 2026-05-02
feat(backup): non-destructive deep-merge restore

tinywind:
## Summary

Add a second restore mode that **merges** a backup into the current library instead of replacing it. The existing replace mode is unchanged and still the default.

## Motivation

Today, restoring a backup wipes everything first, then writes the backup over the empty DB. That's fine for full migrations, but it's lossy in two common cases:

- I read a few chapters on device A, made a backup, kept reading on device B. Restoring B's backup on A throws away A's progress.
- I have novels in my library that aren't in the backup file (added since the backup was taken). Restore deletes them.

The new merge mode keeps both sides:

- **Reading state**: `unread`/`isDownloaded`/`bookmark`/`progress` are OR-merged across both sides; latest `lastRead` wins.
- **Novel/category metadata**: backup wins on conflicts, but rows that exist only in the current DB are preserved.
- **Categories**: deep-merged by name, with id remap so existing novel-category links keep working.

## What's in the PR

| File | Change |
|---|---|
| `src/database/queries/_restoreMergeUtils.ts` | New helpers: `RestoreMode` type, merge policies for novels / chapters / categories. |
| `src/database/queries/NovelQueries.ts` | Existing `restoreNovels` accepts a `mode` param; `merge` path routes through the new utils. |
| `src/database/queries/CategoryQueries.ts` | Same pattern for categories with id remap. |
| `src/services/backup/{local,drive,selfhost}/index.ts` | Plumb `mode` through each backup service. |
| `src/services/ServiceManager.ts` | `LOCAL_RESTORE` / `DRIVE_RESTORE` / `SELF_HOST_RESTORE` accept `data.mode`. |
| `src/screens/settings/SettingsBackupScreen/index.tsx` | New "Restore backup (merge)" entry alongside the existing "Restore backup (replace)". |
| `strings/languages/en/strings.json` + others | New keys: `restoreBackup` / `restoreBackupDesc` / `restoreBackupMerge` / `restoreBackupMergeDesc`. |

## Backwards compatibility

- The default restore path (`mode: 'overwrite'`) goes through the existing code unchanged.
- New `RestoreMode` param is optional everywhere in the service layer; old callers (none in this codebase) keep working.
- No DB schema change.

## Verified scenarios (Galaxy S23, Android 14, release build)

Same baseline backup (`lnreader_backup_2026-05-03_17_17.zip`) used for both checks.

**Replace restore — regression check.** Baseline library had two novels (*A Record of a Mortal's Journey…*, *Child of Light*) plus three categories (`Default` / `Local` / `BackupOnlyCategory`). Backup created, *Child of Light* removed from the library, then restored with the existing replace mode. The library returned to baseline exactly — both novels and all three categories present:

replace mode restores baseline

**Merge restore — preserves novels added after the backup.** Starting from the baseline-restored library, *Ancient Strengthening Technique* was added (so this novel exists locally but not in the backup). Then the same baseline backup was applied with merge mode. All three novels are present afterwards: the two backup novels are restored *and* the locally added one survives:

merge mode preserves new novel

## Test plan

- [x] Replace restore: existing flow still wipes & restores cleanly (regression check). — see capture above
- [x] Merge restore from a backup containing a subset of current library: existing extra novels survive. — see capture above
- [ ] Merge restore where backup has newer `lastRead`: that side wins; older side loses. — *not run on device this round; covered only by the merge utils logic*
- [ ] Merge restore where current DB has newer `lastRead` than backup: current side wins. — *as above*
- [ ] Merge restore: read/downloaded/bookmark flags are OR'd, not overwritten. — *as above*
- [ ] Merge restore: backup categories not present locally get added; existing categories not in backup are kept. — *partially observed in the merge capture (all three pre-existing categories survived)*

## Notes

- Two highest-risk scenarios were verified on a real device this round; the four remaining `lastRead` / per-flag scenarios are unchecked because they need a second device profile (or careful timestamp manipulation) to construct cleanly. Happy to run those if a reviewer wants them before merge.
- This was previously open as #1825 / #1826 but those PRs were closed when the source repo got renamed. Re-opening here against the current upstream `master`.


CD-Z:
You have a test plan section in every pr. Most of the time the points aren't checked, so if you did these tests check them or convert the pr to a draft until you tested it properly and it is ready to merge.

tinywind:
Good call — switched to draft. The boxes were unchecked because those items are still planned scenarios rather than ones I've actually run on a device. I'll work through them on a Galaxy S23 build (replace-restore regression first, then the four merge cases against backups I'll craft locally, then categories), tick each one only after it passes, and flip back to ready for review when the list is fully checked.

tinywind:
Quick follow-up — ran the two highest-risk scenarios on a Galaxy S23 (Android 14, release build) against the same baseline backup:

- **Replace restore** (regression): library returns to baseline exactly after a removed novel.
- **Merge restore** (preservation): a novel added after the backup survives the merge.

Both captures are now embedded in the PR description above. The remaining four boxes (`lastRead` precedence in either direction, OR-merging of read/downloaded/bookmark flags, category union) are still unchecked because they need a second device profile or careful timestamp manipulation to construct cleanly — happy to run any of those before merge if you want them verified on-device too. PR is back to ready for review.

rajarsheechatterjee:
Could you rebase the PR once?

PR #1834, opened 2026-05-03
feat(downloads): make chapter download cooldown configurable

tinywind:
## Summary

Replace the hard-coded 1s sleep between sequential chapter downloads with a single configurable setting under General Settings.

## Why

`downloadChapter.ts` ends every successful chapter write with `sleep(1000)` to be polite to sources. That value is fine as a default but is wrong in two opposite ways:

- **Too aggressive for some sources** — sources that 429 / Cloudflare-block on a tight loop need a longer pause. Users currently work around it by manually pausing the queue.
- **Too slow for sources that allow rapid fetches** — large library downloads (1000+ chapters) burn 15+ minutes of pure idle time without any benefit.

A single global knob is the simplest fix: keep the 1s default so existing installs don't change, but let users turn it up or down.

## What's in the PR

| File | Change |
|---|---|
| `src/hooks/persisted/useSettings.ts` | Add `chapterDownloadCooldownMs?: number` to `AppSettings` and a synchronous `getChapterDownloadCooldownMs()` helper for the background task runner. |
| `src/services/download/downloadChapter.ts` | Replace `sleep(1000)` with `sleep(getChapterDownloadCooldownMs())`. |
| `src/screens/settings/SettingsGeneralScreen/SettingsGeneralScreen.tsx` | New "Chapter download cooldown" entry placed directly above "Disable haptic feedback". Description shows the current value in seconds. |
| `src/screens/settings/SettingsGeneralScreen/modals/DownloadCooldownModal.tsx` | New modal with a numeric input (decimal seconds), Reset, and Ok. Sanitises every keystroke to digits + a single `.` so paste / hardware keyboards can't smuggle in non-numeric characters. Surfaces an error-coloured warning that low values can trigger source rate limits or temporary bans. |
| `strings/languages/en/strings.json` | Four keys under `generalSettingsScreen` (label, description, placeholder, warning). Other locales fall back to English. |

## Backwards compatibility

- Default is `DEFAULT_CHAPTER_DOWNLOAD_COOLDOWN_MS = 1000`, identical to the previous hard-coded value.
- Field is optional in the persisted `AppSettings` shape; existing MMKV blobs upgrade transparently.
- No DB or schema change.

## Screenshots

Captured from a release build on a Galaxy S23 (Android 14).

**List-item placement** — directly above "Disable haptic feedback":

cooldown list item

**Modal** — numeric input, Reset / Ok, and the rate-limit warning under the field:

cooldown modal

## Test plan

- [x] Default install (no override): chapter batch downloads behave exactly like main — 1s between chapters.
- [x] Set cooldown to `0` and run a batch download → no inter-chapter sleep.
- [x] Set cooldown to `2.5`s → roughly 2.5s between completions.
- [x] Reset → value goes back to 1s without leaving the override stale.
- [x] Paste `abc1.5xyz` into the input → reduced to `1.5`.
- [x] Multiple decimal points (`1.2.3`) → reduced to `1.23`.
- [x] Cold-start the app with a saved override and verify the background task runner picks up the value (no rebuild required between settings change and next download).


CD-Z:
The modal should show a warning to the user that low cooldowns can lead to rate limiting or bans from the sources. Also it would be good if you include screenshots of your ui changes.

tinywind:
Both addressed:

- **Warning text**: added in 8a244efb. The modal now shows an error-coloured line under the input — _"Very low cooldowns may trigger rate limiting or temporary IP bans from sources."_
- **Screenshots**: added to the PR description above. Captured from a release build on a Galaxy S23. One shows the new list-item directly above Disable haptic feedback in General Settings, the other shows the modal with the warning visible.

(Screenshots are hosted on a separate `screenshots` branch in my fork to keep this PR's diff strictly on the code change.)

src/screens/settings/SettingsGeneralScreen/modals/DownloadCooldownModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+          {getString('generalSettingsScreen.chapterDownloadCooldown')}
+        </Text>
+        <Text style={[styles.modalDesc, { color: theme.onSurfaceVariant }]}>
+          {getString('generalSettingsScreen.chapterDownloadCooldownDesc')}
+        </Text>
+        <TextInput
+          value={draft}
+          onChangeText={text => setDraft(sanitizeNumericInput(text))}
+          onSubmitEditing={save}
+          keyboardType="decimal-pad"
```
CD-Z, line 90: You should use inputMode instead of keyoardType.
Also use the paper TextInput to keep the style the same with other inputs. I would also let the user enter the time as ms. You can also add an [Affix](https://oss.callstack.com/react-native-paper/docs/components/TextInput/TextInputAffix)

src/screens/settings/SettingsGeneralScreen/modals/DownloadCooldownModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+          style={[
+            styles.input,
+            { color: theme.onSurface, borderColor: theme.outline },
+          ]}
+          autoFocus
+        />
+        <Text style={[styles.warning, { color: theme.error }]}>
+          {getString('generalSettingsScreen.chapterDownloadCooldownWarning')}
+        </Text>
+        <View style={styles.actions}>
```
CD-Z, line 104: Do this the same as in ColorPickerModal

rajarsheechatterjee (approved):

PR #1842, opened 2026-05-09
Better epub import

CD-Z:
Faster import
skip empty chapters / chapters with only title inside
added test
other small fixes

Not tested with lots of epubs

### DIFF
--- a/src/services/epub/import.ts
+++ b/src/services/epub/import.ts
@@ -17,70 +17,19 @@ import NativeFile from '@specs/NativeFile';
 import NativeZipArchive from '@specs/NativeZipArchive';
 import NativeEpub from '@specs/NativeEpub';
 import { eq } from 'drizzle-orm';
+import {
+  decodePath,
+  getExtension,
+  getParentDir,
+  getRelativePath,
+  isDerivedName,
+  isTitleOnlyChapter,
+  normalizePath,
+  resolveAssetPath,
+  shouldSkipUrlRewrite,
+} from './utils';
 
-const decodePath = (path: string) => {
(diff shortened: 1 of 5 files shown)


Palloxin:
finally, did u test and fix all cases I commented in the other old pr?

src/services/epub/import.ts
```diff
@@ -105,6 +105,56 @@ const shouldSkipUrlRewrite = (url: string) => {
   return /^[a-z][a-z0-9+.-]*:/i.test(trimmed);
 };
 
+const stripHtml = (value: string) => {
+  const bodyMatch = value.match(/<body[^>]*>([\s\S]*?)<\/body>/i);
```
Palloxin:
```js
/<body[^>]*>([^]*?)<\/body>/i
```

src/services/epub/import.ts
```diff
@@ -105,6 +105,56 @@ const shouldSkipUrlRewrite = (url: string) => {
...
+  const bodyContent = bodyMatch ? bodyMatch[1] : value;
+  return bodyContent
+    .replace(/<script[^>]*>[\s\S]*?<\/script>/gi, ' ')
```
Palloxin: regex is faster if u replace every `[\s\S]` with `[^]`

Supposing ther egex used is javascript flavour

CD-Z:
This has nothing to do with the old pr. This is for import and the other was for export

Palloxin:
https://github.com/user-attachments/assets/0f4d7ac6-7753-4103-af3e-882a135dfe46

dont forget

PR #1885, opened 2026-07-19
Migration to Expo managed workflow

CD-Z:


rajarsheechatterjee:
What would be the benefits of switching to Expo's managed workflow?

Palloxin:
cm02z9ptzzl8efjikk2z

Edit:
From an internet article, 25 september 2025

CD-Z:
Actually quite a lot.
First off all it is the default for react native these days.
It has more powerful tooling with excellent integration.
It is way easier to upgrade to a new version.
Also was if we want to use it.

CD-Z:
> cm02z9ptzzl8efjikk2z
> 

This looks very outdated

rajarsheechatterjee:
> Actually quite a lot. First off all it is the default for react native these days. It has more powerful tooling with excellent integration. It is way easier to upgrade to a new version. Also was if we want to use it.

If that's the case, it makes sense. Let me read up on it as well.

Issue #1743, opened 2026-02-12
Total time spent


maxplanckthethird:
### Describe your suggested feature

can we add a feature where i can see total time spent reading one specific thing and total time spent reading on the app?

### Other details

it would be good if it was transferable across backups too

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] If this is an issue with a source, I should be opening an issue in the [sources repository](https://github.com/lnreader/lnreader-sources/issues/new/choose).
- [x] I have updated the app to version **[2.0.2](https://github.com/lnreader/lnreader/releases/latest)**.
- [x] I will fill out all of the requested information in this form.

Palloxin:
Third party apps can already measure how much time u use a specific app, android also has that, even without another app it can show u.

i believe u want to know how much long is a novel itself, aka the total word count. You can easly estimate how much u will take to read something with the total word count of the novel. Unfortunately nobody give the word count of web novels, so you will have to harass the aggregators websites in asking for the word-count feature on their database

CD-Z:
I already thought about this with a general stats page upgrade, but I don't know when I have the time to do it.

maxplanckthethird:
> Third party apps can already measure how much time u use a specific app, android also has that, even without another app it can show u.
> 
> i believe u want to know how much long is a novel itself, aka the total word count. You can easly estimate how much u will take to read something with the total word count of the novel. Unfortunately nobody give the word count of web novels, so you will have to harass the aggregators websites in asking for the word-count feature on their database

nah, just tracking how much time i spent reading a specific novel for example, wordcount doesnt really matter

rajarsheechatterjee:
This has been added and will be available in v2.0.4. Closing this issue.

Issue #1168, opened 2024-07-29
Time spent on reading tracker


OMNITHORN:
### Describe your suggested feature

It would be extremely helpful if there was a tracker for time spent reading webnovels in Statistics

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with a source, I should be opening an issue in the [sources repository](https://github.com/LNReader/lnreader-sources/issues/new/choose).
- [X] I have updated the app to version **[2.0.0](https://github.com/LNReader/lnreader/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

PR #1886, opened 2026-07-20
feat: implement time-tracking and statistics

Ecasept:
### Feature
Tracks how long the user spends reading each individual chapter in non-incognito mode, and shows the data inside the statistics screen.
- New `StatsCard` showing total time spent reading (including novels not in the library)
- New section showing the top 10 novels/categories by time spent, with ability to switch between modes

### Implementation
- `useTimeTracking` stores the timestamp of when the current chapter was opened
- When the current chapter changes (because the user navigated to another chapter), the useEffect ends (because the user navigated away from the reader), when the app state changes, and periodically every minute, we write add the elapsed time to a new database column on the current chapter, and reset the elapsed time.
- The stats screen sums up the time spent in total and for each novel/category, and displays them
- The new `useTimeTracking` hook and changes inside the chapter actions, stats queries and chapter queries are fully covered by tests
- The chapter schema in the database received a new `timeSpent` column and was migrated
- I removed some redundant code from a previous PR

Any feedback is welcome

Addresses: #1743 #1168

### Screenshots
**Sorted by Novel:**
Screenshot_20260720-182233
**Sorted by Category:**
Screenshot_20260720-182238
**Total Time:**
Screenshot_20260720-182228

### DIFF
--- a/drizzle/20260719143427_long_moondragon/migration.sql
+++ b/drizzle/20260719143427_long_moondragon/migration.sql
@@ -1,0 +1,1 @@
+ALTER TABLE `Chapter` ADD `timeSpent` integer DEFAULT 0;
--- a/drizzle/20260719143427_long_moondragon/snapshot.json
+++ b/drizzle/20260719143427_long_moondragon/snapshot.json
@@ -1,0 +1,644 @@
+{
+  "version": "7",
+  "dialect": "sqlite",
+  "id": "498c0b85-a3f4-413f-8fed-7914c3ba15d9",
+  "prevIds": [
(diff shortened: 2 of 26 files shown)


Palloxin:
An option to turn it off? In the sense to not make the thing count the time.

Basically to disable it as if this pr feature doesnt exist, no time counting whatsoever even when incognito mode is off.

Ecasept:
Yeah, you're right that would make sense. I'll get to it by tomorrow

rajarsheechatterjee:
I'm merging this since I'm planning a few other changes that might cause conflicts. You can raise a separate PR for the setting.

Thanks for the contribution! 🚀

PR #1919, opened 2026-07-27
Enhance stats UI with genre screen, charts, and normalization

CD-Z:
Screenshot_1785420153
Screenshot_1785420158
Screenshot_1785420166
Screenshot_1785421308
Screenshot_1785421418
Screenshot_1785232364


rajarsheechatterjee:
Do you think there's a better way to display "Titles in Library", "Sources", and "Total Time Spent"?

CD-Z:
The titles in library is already displayed in the center of the pie chart. So this can be removed.
Then I thought about adding a second pie chart with the sources in the middle and the distribution around or making the pie chart pressable to switch between both charts.
With the total time spent I'm not sure. Maybe stretching the card wider and giving it primary as background.

rajarsheechatterjee:
> The titles in library is already displayed in the center of the pie chart. So this can be removed. Then I thought about adding a second pie chart with the sources in the middle and the distribution around or making the pie chart pressable to switch between both charts. With the total time spent I'm not sure. Maybe stretching the card wider and giving it primary as background.

Then for now this should be fine. Let me also think about the design.

CD-Z:
I also forgot to show the genre distribution properly in the screenshots. I will do it later as I'm currently not at my pc.

CD-Z:
@rajarsheechatterjee I updated the images now.

rajarsheechatterjee:
> @rajarsheechatterjee I updated the images now.

Looks good to me.

rajarsheechatterjee (approved):

PR #1923, opened 2026-07-28
Ignore Notify Discord if not on the main repo

CD-Z:


rajarsheechatterjee (approved):

Palloxin:
is possible to make app build even without the tracked id client?

if (lnreader/lnreader) => deleted client id etc.  So other users can very esly build the apk without knowing how to code, or github stuff, just by forking and triggering workflow

Palloxin:
to by pass this limitation
Screenmj

CD-Z:
No, this is harder. Just go to https://github.com/palloxin/lnreader-clone/settings/variables/actions
and input
image

Palloxin:
Is there really no other solution?

like, cool, i fix it for myself, but what about other testers?

PR #1933, opened 2026-07-29
Activate react compiler

CD-Z:
Reactivated react compiler and fix most react compiler errors.


src/screens/BrowseSourceScreen/useBrowseSource.ts
```diff
@@ -4,10 +4,20 @@ import { NovelItem } from '@plugins/types';
 import { getPlugin } from '@plugins/pluginManager';
 import { FilterToValues, Filters } from '@plugins/types/filterTypes';
 
+function getOrThrowPlugin(pluginId: string) {
+  'no use memo';
```
rajarsheechatterjee, line 8: Shouldn't this be `use no memo`?

src/hooks/common/useGithubUpdateChecker.ts
```diff
@@ -26,28 +26,48 @@ const getReleaseDownloadUrl = async (
...
+    );
   }
+  for (const abi of supportedAbis) {
+    const matchingAsset = apkAssets.find(asset =>
+      asset.name?.endsWith(`-${abi}.apk`),
+    );
 
-  return (
-    universalAsset?.browser_download_url || apkAssets[0]?.browser_download_url
-  );
```
rajarsheechatterjee, line 48: The universal APK fallback should still remain after the loop, not only in the exception handler.

PR #1938, opened 2026-07-30
fix(backup): include all chapters beyond the 1000-row UI limit

aliomar139:
## Summary

Fixes incomplete backups for novels containing more than 1,000 chapters.

- Adds a dedicated, unbounded chapter query for backup creation.
- Keeps the existing 1,000-row limit for interactive screens.
- Updates backup generation to use the complete chapter query.
- Adds a regression test covering a novel with 1,001 chapters.

## Problem

Backup generation reused `getNovelChapters()`, which defaults to a limit of 1,000 rows.

As a result, novels with more than 1,000 chapters had incomplete chapter metadata in the backup, even when their downloaded chapter files were present. Restoring such a backup could:

- Display only 1,000 available or downloaded chapters.
- Lose the association between downloaded files and chapter records.
- Make later library updates treat existing chapters as new chapters.
- Produce duplicate or mismatched downloads.

This affected any long novel, including Shadow Slave, Reverend Insanity, LOTM 2, and Pursuit of the Truth.

## Solution

A separate `getAllNovelChaptersForBackup()` query now returns every chapter for a novel without imposing the UI pagination limit.

Backup creation uses this query, while normal chapter screens continue using the existing limited query for performance.

## Testing

- Added a regression test inserting 1,001 chapters.
- Verifies the normal UI query returns 1,000 chapters.
- Verifies the backup query returns all 1,001 chapters.
- TypeScript type-check passed.
- Formatting and pre-commit lint checks passed.

## Compatibility

- No database migration is required.
- Existing UI pagination behavior is unchanged.
- New backups will contain complete chapter metadata.
- Existing incomplete backups are not automatically repaired.


scripts/repair-capped-backup.ps1
rajarsheechatterjee, line 1: Is this being used anywhere?

pnpm-lock.yaml
rajarsheechatterjee, line 1: Is this expected?

.gitignore
```diff
@@ -114,3 +114,7 @@ release-artifacts/
 !plugins/android/
 ios/
 !modules/**/ios/
+lnreader_backup_2026-07-27_09_35.zip
+lnreader_backup_2026-07-27-deduplicated.zip
+lnreader_backup_2026-07-27-fixed.zip
+lnreader_backup_2026-07-27-ready-to-restore.zip
```
rajarsheechatterjee:
```suggestion
```

scripts/repair-capped-backup.ps1
aliomar139: It was a one-off recovery utility and wasn’t used by the app, build, or CI. I’ve removed it from the PR.

pnpm-lock.yaml
aliomar139: It was an unintended unrelated change. This fix doesn’t modify dependencies, so I restored the lockfile to the upstream version.

rajarsheechatterjee:
Could you fix the failing tests? I'll merge this after that.

aliomar139:
Fixed the failing Jest test , the backup test was mocking the old function name

rajarsheechatterjee (approved):

PR #1971, opened 2026-08-08
Fix novel screen scroll lag, source pagination stalls, and skeleton overlap

### LINKED ISSUE #1950: Heavy Lag upon opening new novels
### Steps to reproduce

This was a bit hard to recreate but i did it. So first, 
1.open any plugin
2.Open any novel

### Expected behavior

It'll lag for a bit after you novel novels but will go away soon, i guess this happens coz it's loading the chapters or something? Hmm.. beta3 didn't have this problem tho

### Actual behavior

So most of the time, the lag will go away in just 1/2secs of opening the novel, BUT, sometimes, that lag stays, it'll continue to lag even when you leave the laggy novel and just browse the plugin. But when you leave plugin fully, lag goes away, but when you open plugin again, lag comes back. Only way to fix this is to restart app. Here's a video, use the gofile link for better quality video( the continuous lag starts at the end of video, on novel "The Great Ming in the Box", so notice it carefully)

[Record_2026-08-03-23-45-57.mp4.zip](https://github.com/user-attachments/files/30674655/Record_2026-08-03-23-45-57.mp4.zip)

https://gofile.io/d/yeVlIb

### LNReader version

Preview

### Android version

10

### Device

Realme 7 pro

### Other details

_No response_

### Acknowledgements

- [x] If this is an issue with a plugin, I should be opening an issue in the [plugins repository](https://github.com/lnreader/lnreader-plugins/issues/new/choose).
- [x] I have updated the app to version **[2.1.2](https://github.com/lnreader/lnreader/releases/latest)**.

### LINKED ISSUE #1951: Error upon novel opening
### Steps to reproduce

1. Add any ovel to library 
2. Open any chapter
3. Come back to library 
4. Click on that novel again
5. Error pops up

### Expected behavior

Should open normally 

### Actual behavior

https://github.com/user-attachments/assets/205bcb42-d209-448c-8f25-c2b9e511809c

### LNReader version

Preview

### Android version

10

### Device

Realme 7 pro

### Other details

Plugin doesn't matter, happening with all plugins

### Acknowledgements

- [x] If this is an issue with a plugin, I should be opening an issue in the [plugins repository](https://github.com/lnreader/lnreader-plugins/issues/new/choose).
- [x] I have updated the app to version **[2.1.2](https://github.com/lnreader/lnreader/releases/latest)**.

CD-Z:
## Summary

Three issues fixed on Android:

1. **The novel screen scrolls at ~20fps after browsing a few uncached novels**, persisting app-wide until restart.
2. **The source browser stops loading further pages** permanently after a transient fetch error.
3. **The loading skeleton at the end of the source list overlaps.**

The lag's root cause: the `AnimatedLegendList` reanimated integration leaked stale animated nodes from recycled/unmounted views. Every scroll event pushed prop updates for those views, Fabric threw per-tag, and reanimated logged a **full stack trace on the main thread for each failure** — ~40ms/frame at ~160 stale tags per event. The nodes accumulated per screen mount (hence "after ~6 opens"), and only a restart cleared them.

## Fixes

- **Novel screen — drop the AnimatedLegendList** (`NovelScreenList.tsx`): replaced with the plain `LegendList` + a plain `onScroll` handler writing the same `scrollOffset` shared value. The scroll offset was only consumed by the existing `useAnimatedReaction` (header fade, FAB, scroll-to-top) — verified still working. Removes the stale-node source entirely.
- **Source browser — kill the perpetual fetch loop and permanent pagination death** (`BrowseSourceScreen.tsx`, `useBrowseSource.ts`): `onEndReached` on a short list re-fired forever on sources that never return an empty page (measured: htmlparser2 hot on every frame). Replaced with an `onScroll` distance gate that fetches only while the user approaches the end and re-arms after appends. Transient fetch errors no longer set `hasNextPage = false`; the failed page is retried on the next approach instead of skipped.
- **Novel screen — defer batched chapter reads out of the scroll path** (`novelStore.chapterActions.ts`): the 1000-row sync DB reads triggered by `onEndReached` are now deferred to idle (`runWhenIdle`), removing boundary stalls on >1000-chapter novels.
- **Skeleton footer — fix the overlapping trailing row** (`NovelList.tsx`, `SourceScreenSkeletonLoading.tsx`): the `completeRow: 2` item rendered a full grid inside one 1/3-width cell; it is now a proper full-width `ListFooterComponent` row.
- **NovelContextProvider hardening** (`NovelContext.tsx`): null-safe route-param handling (fixes `Cannot read property 'path' of undefined` when opening a library novel after a plugin open), plus a deferred `gc()` on store unmount to reclaim the ~5MB/op heap ratchet (Hermes' GC only runs on allocation pressure; measured 31.2 → 26.4MB live).

## Verification (Android emulator, debug build, Novel Fire)

| Scenario | Before | After |
|---|---|---|
| Novel scroll after 6 uncached opens | 65% jank, 53ms median; 31,967 reanimated failure logs/session | 2.3% jank, 16/21/24ms p50/p90/p95, 0 failures |
| Live heap after 6 opens | ~58MB, never reclaimed | 28MB |
| Source pagination | stops forever after one error; infinite loop otherwise | loads on approach, retries failed pages |
| Skeleton footer | overlapping grid in one cell | single full-width row |

`tsc` clean; 105 relevant tests pass; prettier clean. (The pre-existing `NovelQueries` cover-unlink test failure is untouched.)

## Files changed

- `src/screens/novel/components/NovelScreenList.tsx`
- `src/screens/BrowseSourceScreen/BrowseSourceScreen.tsx`
- `src/screens/BrowseSourceScreen/useBrowseSource.ts`
- `src/hooks/persisted/useNovel/store/novelStore.chapterActions.ts`
- `src/hooks/persisted/useNovel/__tests__/novelStore.chapterActions.test.ts`
- `src/screens/novel/NovelContext.tsx`
- `src/components/NovelList.tsx`
- `src/components/__tests__/NovelList.test.ts`
- `src/screens/browse/loadingAnimation/SourceScreenSkeletonLoading.tsx`

fixes #1950, #1951


CD-Z:
Made a missclick. Accidentally pushed this branch to upstream instead of origin

rajarsheechatterjee (approved):

Issue #1975, opened 2026-08-10
Bug: TTS continues playing during incoming and outgoing phone calls


alvannoah:
### Steps to reproduce

1. Open a novel in LNReader.
2. Start Text-to-Speech playback.
3. While TTS is playing, receive an incoming phone call.
4. Answer the call.
Observe that LNReader's TTS continues playing.

The same behavior can be observed when initiating an outgoing phone call while TTS is playing.

### Expected behavior

TTS should automatically pause when an incoming or outgoing phone call becomes active.

After the call ends, TTS should resume playback from where it was paused.

### Actual behavior

TTS continues playing during the phone call.

### LNReader version

2.1.2

### Android version

Android 15

### Device

Infinix Hot 50 pro+

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] If this is an issue with a plugin, I should be opening an issue in the [plugins repository](https://github.com/lnreader/lnreader-plugins/issues/new/choose).
- [x] I have updated the app to version **[2.1.2](https://github.com/lnreader/lnreader/releases/latest)**.
- [x] I will fill out all of the requested information in this form.

PR #1976, opened 2026-08-10
fix: TTS continues playing during incoming and outgoing phone calls (Closes issue #1975)

alvannoah:
Implement audio focus handling for TTS playback on Android, including support for request and abandonment during playback and transient interruptions.


alvannoah:
Hello, this is my first contribution to LNReader.
I request for guidance in case there are any contribution policies violated.
Thank you.

PR #2007, opened 2026-08-22
feat(history): bulk-remove novels from Library via long-press select (#1874)

### LINKED ISSUE #1874: Selective novel chapter removal
### Describe your suggested feature

Can't y'all optimize the history section to just hold and select a novel and it automatically deletes all the chapters from the library as well as downloaded chapters.

It's all a pain in ass when we have to do it all manually, i think it can be done with just a few lines of code.

Second feature request is deleting unwanted notification from the update section, for example if i don't want to see a novel in the updated section, then i can simply delete the update notification.

### Other details

_No response_

### Acknowledgements

- [x] If this is an issue with a source, I should be opening an issue in the [sources repository](https://github.com/lnreader/lnreader-sources/issues/new/choose).
- [x] I have updated the app to version **[2.0.3](https://github.com/lnreader/lnreader/releases/latest)**.

RibatTRW:
# Bulk-remove novels from History (hold-and-select) — #1874

## What this does

Long-press any novel card in **History** and the reader enters multi-select
mode — the same pattern Library already uses. Pick one novel or a hundred,
hit **Remove**, confirm, and every selected novel is purged completely:
queued downloads for those novels are cancelled, downloaded chapter files
are deleted, the novel is removed from Library (when it's in Library), and
its reading history is cleared. A confirm dialog shows exact counts before
anything happens, and a toast reports the outcome afterwards.

Closes #1874.

## How it behaves

- **Selection:** long-press a History card → checkboxes + an action bar
  (Remove / cancel). Search and section headers keep working in select mode.
- **Confirm dialog:** shows the computed counts — novels selected, how many
  will leave Library, and how many downloaded chapters will be deleted
  (live counts fetched when the dialog opens, not stale row data).
- **Purge order:** cancel that novel's queued
  downloads first, then delete its files, then one batched Library removal,
  and **history clears last** — only for novels whose purge reached that
  stage. If the Library step fails, affected novels keep their history
  (the irreversible step never runs on a half-failed purge).
- **Error isolation:** one failing novel never aborts the rest. Failures are
  named individually in the outcome toast (`novelName: error`).
- **Queued-download safety:** only the selected novels' queued downloads are
  cancelled — never the whole queue. Tasks queued by an older app version
  (no novel id) are deliberately left running and surfaced via toast instead
  of being killed blind.
- **Everything is localized** (en strings + generated types); no hardcoded
  copy in the touched paths.

## Verification

- `pnpm run test:rn`: **79/79 suites, 381/381 tests PASS** at tip `744a1776`
  (independently re-run by review).
- `tsc --noEmit` exit 0; prettier + eslint clean on all delta files.
- Test floor: a 9-test pure orchestration suite pinning the purge
  order-of-operations, atomicity and error isolation, plus 3 direct
  `cancelForNovels` return-contract tests — all with module-level mocks
  per the repo's TESTING.md.
- Reviewed across three structured review rounds; every review blocker
  was resolved with committed work and the final state was independently
  countersigned before this PR was filed.

## Notes

- **Trackers are untouched** — tracker bindings are server-side and this
  feature does not interact with them.
- **Disclosure:** the queued-download pre-flight only fires when a selected
  novel actually has something in the live queue; legacy-task surfacing
  therefore only appears in that case.
- **AI attribution:** generated with Hermes (builder-bob agent), reviewed by
  a human; `Co-authored-by: Hermes builder-bob <builder-bob@hermes.local>`
  on every commit.
- Base: branched directly off upstream commit `7e20f7dc` (the fork's
  own master ref had drifted behind upstream at the time); range
  `b2bd2ba6..744a1776`.


RibatTRW:
Follow-up commit `c54ece0` fixes two defects found in a post-file review pass:

**1. Confirm-dialog counts were per history row, not per novel.** History rows are per-chapter, so selecting one novel with N read chapters made the dialog report N novels and multiply its downloaded-chapter count by N. `HistoryScreen` now dedupes the selection by `novelId` before computing `novelCount`, `libraryNovelCount`, and the live chapter count. (The purge itself already deduped — only the dialog lied.)

**2. Double toast on confirm.** The purge wired the real `removeNovelsFromLibrary`, which fires its own toast, and the hook then fired `purgeSuccess`. `NovelQueries` now splits out `removeNovelsFromLibrarySilent` (DB write only) for the purge path; the toasting wrapper keeps existing Library behavior unchanged.

Added two regression tests at the screen seam (long-press → dialog → confirm), driving the real components and the real `NovelQueries` module — both fail on `744a177` and pass on `c54ece0`. Full `test:rn` shows no other deltas; `tsc --noEmit` and eslint clean.

PR #2009, opened 2026-08-23
feat(reader): RSVP (speed) reading mode (#1576)

RibatTRW:
# RSVP (speed) reading mode — #1576

## What this does

A new **RSVP tab** in the reader's bottom sheet starts speed-reading mode:
the chapter flashes word-by-word or in small chunks at a fixed position on
screen, at a pace you control from the reader's bottom-sheet RSVP tab —
pause, resume, exit, WPM slider, chunk size — even mid-session. Exiting
lands you back in normal reading scrolled to the exact element you last
read.

Implements the RSVP reading mode requested in discussion #1576.

## How it works

- **One text pipeline:** the flash queue comes from the SAME DOM walk TTS
  uses (the shared collector in `core.js`) — there is exactly one traversal
  implementation; RSVP consumes its output rather than re-walking.
- **Cadence runs inside the WebView:** base interval is 60000/WPM with
  punctuation-proportional pauses (comma ≈ ×1.5, sentence ≈ ×2.5,
  paragraph break ×3). Words-per-minute (WPM) changes apply live. The
  native side hears position updates only on pause/exit/save ticks — never
  per-flash.
- **Chunking:** 1–3 words per flash (user-selectable). Tokens longer than
  12 characters are split across consecutive flashes with ellipsis
  continuation. Non-text nodes flash an honest "[image]" placeholder so
  progress stays true.
- **Progress:** entering speed-read mode starts from the chapter beginning;
  exiting saves the exact element so normal reading resumes where the
  speed-read stopped — through the existing auto-save path.
- **Settings:** words-per-minute (WPM, default 250) and chunk size
  (default 1) persist across restarts via MMKV. Note these settings are not
  included in backups today.
- **Everything localized** via en strings + generated types.

## Verification

- `pnpm run test:rn`: **81/81 suites, 400/400 tests PASS** at tip.
- `tsc --noEmit` exit 0; prettier + eslint clean on all delta files.
- Test floor: chunk splitter and pause calculator are pure functions with
  node-env tests (red-first documented); a bridge contract suite loads the
  real `core.js` in a minimal DOM stub and pins the surface plus the
  one-message-per-pause/exit rule; settings hook tests cover defaults,
  bounds clamping (150–800 WPM), and multi-patch composition.

## Notes

- **Maintenance disclosure:** the WebView-side chunk builder mirrors the
  native `chunkSplitter`/`pauseCalculator` logic across the
  JavaScript/TypeScript boundary (WebView code cannot import TS modules).
  The contract tests on both sides pin identical behavior — changes to
  either must keep them in step.
- **Trackers are untouched** by this feature.
- **AI attribution:** generated with Hermes (builder-bob agent), reviewed by
  a human; `Co-authored-by: Hermes builder-bob <builder-bob@hermes.local>`
  on every commit.
- Base: upstream tip `cb1a5d9e` (v2.1.3); branch
  `feature/1576-rsvp-reading`, range `cb1a5d9e..9976f873`.

### DIFF
--- a/assets/reader/js/core.js
+++ b/assets/reader/js/core.js
@@ -166,6 +166,25 @@ window.reader = new (function () {
     console.warn = console.log;
     console.error = console.log;
   }
+  /**
+   * #1576 AC5: translate an RSVP element index into the chapter-progress
+   * percentage the existing autoSave path understands, and scroll the
+   * real DOM to that element so exit lands on the exact last-read spot.
+   */
+  this.saveRsvpPosition = elementIndex => {
+    const elements = collectReadableEntries(this.chapterElement);
+    const clamped = Math.max(0, Math.min(elementIndex, elements.length - 1));
+    const target = elements[clamped];
+    if (target) this.scrollToElement(target.element);
+    if (elements.length > 0) {
+      const pct = Math.round(((clamped + 1) / elements.length) * 100);
+      if (pct > (this.chapter.progress || 0)) {
+        this.chapter.progress = pct;
+        this.post({ type: 'save', data: pct });
+      }
+    }
+  };
+
   // end reader
 })();
 
--- a/assets/reader/js/core.js
+++ b/assets/reader/js/core.js
@@ -1027,3 +1027,209 @@ window.addEventListener('load', () => {
     }
   });
 })();
+
+// --- RSVP bridge (#1576) -------------------------------------------------
+// window.rsvp: speed-reader mode reusing the SAME collectReadableEntries
+// output as TTS (no second DOM walk). The cadence timer lives here in the
+// WebView; the native side only hears 'rsvp-position' on pause/exit/save —
+// never per-flash. Chunking math mirrors the pure functions tested natively
+// (chunkSplitter.ts / pauseCalculator.ts): chunks of 1–3 words, >12-char
+// tokens split at 12 chars with continuation flags, punctuation pauses.
+window.rsvp = new (function () {
+  this.active = false;
+  this.paused = false;
+  this.entries = []; // [{ element, text }] from the shared collector
+  this.chunks = []; // [{ text, elementIndex, continuesNext, continuationOfPrev }]
+  this.chunkIndex = 0;
+  this.wpm = 250;
+  this.chunkSize = 1;
+  this.timerId = null;
+
+  const FLASH_OVERLAY_ID = 'RSVP-Overlay';
+
+  const buildChunks = () => {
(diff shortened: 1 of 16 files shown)


RibatTRW:
**Status update — and an honest correction to this PR.**

An external review of this PR found a fatal bug in the RSVP controls as originally submitted: `RsvpTab`'s `send()` built its injected JavaScript with optional chaining but **no call parentheses**, so every button/slider produced syntactically invalid JS (`SyntaxError`) that the WebView discarded. In practice: **RSVP speed-reading could not be started from the UI at all** in the original head of this PR.

**Fixed in `e57b1caf`** (pushed to this branch): the injection now wraps the argument in call parens, matching the repo's existing TTS injection idiom. A 162-line regression suite (`RsvpTab.test.tsx`) now parses every captured script exactly as a WebView would — a syntactically invalid injection fails the test instead of silently no-oping on device — and asserts which `window.rsvp` methods each control actually invokes.

Credit where due: the bug was found by an independent external review session, independently verified by our QA (semantic proof via `new Function()` on the literal injected strings, red→green reproduced digit-for-digit), then fixed and pushed. The original submission's test suite was green because it never exercised the injection seam — that seam is now covered.

Apologies for shipping the broken form initially; the regression tests exist precisely so this class of failure can't slip through again.

RibatTRW:
## Cross-reference / merge-order note (verified)

- Related open PR: #2014 "refactor(reader): WebView bridge extraction — typed namespace slices with parity tests" (`refactor/reader-bridge`). Both are **siblings** from upstream master @ `cb1a5d9e` (merge-base verified; neither contains the other's commits).
- **Overlap**: exactly one shared file — `src/screens/reader/components/WebViewReader.tsx`, with disjoint diff regions (this PR: toast import + customJS/customCSS + WebView JSX restructure; #2014: bridge imports + adjacent-chapter script builder + component head).
- **Order**: **no required order** — either merges first; the other needs at most a trivial rebase. Expected conflict surface, if any: `WebViewReader.tsx` only.
- **rsvp dependency**: #2014 intentionally **excludes** the rsvp slice (no rsvp references at all — trackable reason: card t_38ff5fae). Once this PR's `window.rsvp` surface is on master, the rsvp bridge follow-up extracts this PR's rsvp wiring into a typed module following the #2014 pattern. So this PR is a **prerequisite** only for that follow-up, not for #2014 itself.
- After the second merge, run `pnpm test:rn` (both PRs' suites share the real `assets/reader/js/core.js` via #2014's parity harness).

Issue #1997, opened 2026-08-20
[[CODE FEATURE]] Make a test simulation for EPUBs, Import/Export to avoid fallbacks


Palloxin:
### Describe your suggested feature

I wish this repo had a testing ground to avoid fallbacks with IMPORT/EXPORT of epubs. Offline, not github workflow (I assume it is too expensive)

An archive of epubs, reduced to the *minimum* (minimum if u want to add the test ground in thus repo) possible enough to emulate their specific use-case. Tho I am not asking to add this test on WORKFLOW cuz too expensive I assume. 

An offline test ground, and when new Import/Export epub bugs pop up, their cases must be added to the test ground

### Other details

I am not sure if it is clear enough.
I have this js "test ground" to check if my custom js doesnt "fall back" with every new change I add

https://github.com/user-attachments/assets/264ffa32-da61-43cb-9a15-fd2971ff8079

I have seen lately conitnuous chamges of codes and there are a lotmof.fallbacks.

-Epubs that imported before, dont now

-A 50mb epub imported weights 94mb inside the related ``Android`` folder (a clear case of image duplication)

-Epubs that imported in 2 minutes on beta3, now takes 5m 

-Images thay get duplicated, or simply skipped and not added
-Export bugs etc.
so on and so corn

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] If this is an issue with a source, I should be opening an issue in the [sources repository](https://github.com/lnreader/lnreader-sources/issues/new/choose).
- [x] I have updated the app to version **[2.1.2](https://github.com/lnreader/lnreader/releases/latest)**.
- [x] I will fill out all of the requested information in this form.

PR #2011, opened 2026-08-23
test(epub): regression-test ground for EPUB import — fixtures, budgets, silent-catch pin (#1997)

RibatTRW:
Adds a regression-test ground for the EPUB import/export pipeline (#1997): the import flow previously had zero tests, and its silent error handling meant broken imports showed a toast while progress quietly completed — so changes to the parsing path could silently degrade without anything failing.

**What this adds**
- Pure helpers (`decodePath`, chapter-name fallback) extracted from `importEpub` into `helpers.ts`, each pinned by node tests — zero behavior change.
- A procedural fixture generator: test EPUBs are described as plain config data and built at runtime; no binary fixtures are committed to the repo.
- Orchestration tests for the real `import.ts` covering: successful import (novel flagged as local + in library, category assigned), name fallbacks for unnamed novels and chapters, an exact operation-count budget on file moves (catches the "50 MB import becomes 94 MB" duplication class deterministically), and the current silent-error behavior pinned explicitly — a follow-up change will make failures loud by flipping exactly those assertions.
- `CASES.md`: a living manifest mapping known bug symptoms to their tests, plus an intake protocol so future EPUB bugs land as failing tests before fixes.

**Honest boundary:** the actual EPUB parser is native code that can't run in Jest. These tests cover everything around it — orchestration, file handling, database writes, fallbacks — with the parser mocked at exactly one seam. The boundary is documented in the test file itself.

**Testing**
- 80/80 test suites, 384/384 tests green; type-check, lint, and formatting clean.
- New tests fail against the pre-fix behavior where applicable (red-first documented in commit messages).

Trackers are untouched. Test-only settings/fixtures are not included in backups.

Refs #1997

Generated with Hermes (builder-bob agent).

Co-authored-by: Hermes builder-bob <builder-bob@hermes.local>

### DIFF
--- a/src/services/epub/CASES.md
+++ b/src/services/epub/CASES.md
@@ -6,22 +6,23 @@ with a test in `importEpub.test.ts` (or a new sibling file) before fixing.
 
 ## Pinned cases (covered by tests)
 
-| Symptom                                                            | Fixture / test                                                                |
-| ------------------------------------------------------------------ | ----------------------------------------------------------------------------- |
-| Novel imported but not in library / missing local flags            | AC3 happy path — `importEpub.test.ts` "inserts the novel local+inLibrary"     |
-| Category not auto-assigned on local import                         | AC3 happy path — same test, `updateNovelCategoryById(100, [2])`               |
-| Unnamed novel imported under a garbage internal name               | Name-fallback test — "derives the novel name from the filename"               |
-| 50 MB EPUB balloons to ~94 MB after import                         | AC4 moveFile budget — "calls moveFile exactly once per existing source image" |
-| Missing image aborts the whole import                              | AC4 skip test — "skips moveFile for sources that do not exist"                |
-| Import failure invisible: toast fires, progress completes silently | AC5 silent-catch pin — "toasts and still terminates progress at 1"            |
+| Symptom                                                                     | Fixture / test                                                                                           |
+| --------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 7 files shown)


RibatTRW:
**Note for maintainers: merge order for the two EPUB-related PRs**

PR #2011 (EPUB test-ground) and PR #2013 (EPUB port refactor) touch overlapping files in `src/services/epub/`. They were developed sequentially and **#2013's branch includes #2011's commits**, so:

- **Recommended merge order: #2011 first, then #2013.**
- If #2013 merges first, #2011 will likely show conflicts or appear already-partially-merged, since its changes are a strict subset of #2013's history.
- Both were reviewed against this dependency: #2013's tests build on the test-ground budgets established in #2011.

Happy to rebase whichever PR remains after the first merge if that eases review.

RibatTRW:
**Correction to my earlier merge-order note** (supersedes the previous comment):

Git verification shows these two PRs are **independent sibling branches**, both cut from the same base (`cb1a5d9e`) — #2013 does *not* contain #2011's commits as I previously stated. They touch 3 overlapping files (`import.ts`, `fixtures/generator.ts`, `CASES.md`).

- **Sequential merge (#2011 → #2013) is the clean path**: no rebase needed; #2013 builds on #2011's test foundation.
- **Reverse order (#2013 first)**: #2011 would need one rebase and its import-side test updates rewritten against the port interface, since `import.ts` and its orchestration tests would already be deleted.

Either order works with intervention; sequential is friction-free. Apologies for the earlier imprecision — the dependency direction matters for whoever merges second.

PR #2013, opened 2026-08-24
refactor(epub): deep import/export port with typed errors and progress phases

RibatTRW:
Restructures the EPUB import/export code into a single deep module with typed errors (#1997 architecture follow-up)

The EPUB import and export paths previously lived in two entry-point modules where any error was caught, shown as a toast, and swallowed. That made regressions invisible: broken imports looked like completed ones, and image files could be silently duplicated (a 50 MB book growing to ~94 MB) without anything failing.

**What changed**
- Import and export are now two functions (`importNovel`, `exportNovel`) that return a structured result: either success data or a typed list of errors (parse failure, missing file, corrupt archive, partial image moves with the exact file list, database write failure). No toasts inside the module; screens decide what to show.
- Progress reporting uses named phases (copy, extract, parse, db, images) instead of loose strings.
- Export assembly (chapter paths, metadata, file naming) moved inside the module. Callers pass intent only: which novel, destination, filename.
- The old entry points are deleted, not deprecated. TypeScript enforces that nothing can reach them.
- Tests cross only the public interface: happy paths, every error kind, phase ordering, and an operation-count budget asserting each existing source file is moved exactly once (this is what makes the duplication bug class fail loudly). A manifest (`CASES.md`) maps known symptoms to tests for future bugs.

**Testing**: full suite green (76 suites / 375 tests), type-check, lint, and formatting clean.

Refs #1997

Generated with Hermes (builder-bob agent).

Co-authored-by: Hermes builder-bob <builder-bob@hermes.local>


RibatTRW:
**Note for maintainers: merge order for the two EPUB-related PRs**

PR #2011 (EPUB test-ground) and PR #2013 (EPUB port refactor) touch overlapping files in `src/services/epub/`. They were developed sequentially and **#2013's branch includes #2011's commits**, so:

- **Recommended merge order: #2011 first, then #2013.**
- If #2013 merges first, #2011 will likely show conflicts or appear already-partially-merged, since its changes are a strict subset of #2013's history.
- Both were reviewed against this dependency: #2013's tests build on the test-ground budgets established in #2011.

Happy to rebase whichever PR remains after the first merge if that eases review.

RibatTRW:
**Correction to my earlier merge-order note** (supersedes the previous comment):

Git verification shows these two PRs are **independent sibling branches**, both cut from the same base (`cb1a5d9e`) — #2013 does *not* contain #2011's commits as I previously stated. They touch 3 overlapping files (`import.ts`, `fixtures/generator.ts`, `CASES.md`).

- **Sequential merge (#2011 → #2013) is the clean path**: no rebase needed; #2013 builds on #2011's test foundation.
- **Reverse order (#2013 first)**: #2011 would need one rebase and its import-side test updates rewritten against the port interface, since `import.ts` and its orchestration tests would already be deleted.

Either order works with intervention; sequential is friction-free. Apologies for the earlier imprecision — the dependency direction matters for whoever merges second.

